PR #6, opened 2017-07-29
Fixed all warnings / Enum errors, made other small changes

Xavia91:


Xavia91:
ignore the reverts^^ eclipse would not let me revert some project wide changes so I had to cheat 😒

Xavia91:
reverted some fixes that have already been fixed in the original branch and caused conflicts.

PR #28, opened 2017-08-16
hackfix clothing colors in encyclopedia - needs a proper fix later

Pimvgd:
This doubles the height if there are more than 15 colors for a piece of clothing.

In the future, you'd want to have a solution that dynamically scales instead of just doubling the height (mainly because this will break for 31 colours), but it should be fine for now.


Rfpnj:
I do not see an issue with this being a long term solution to the issue, having more than 30 colors for clothing is extreme for any game.

PR #33, opened 2017-08-20
Allow assigning shift and control keybindings

Pimvgd:


src/com/base/game/KeyCodeWithModifiers.java
```diff
@@ -0,0 +1,120 @@
...
+
+import com.base.utils.Util;
+
+import javafx.scene.input.KeyCode;
+import javafx.scene.input.KeyEvent;
+
+/**
+ * @since 0.1.83
+ * @version 0.1.83
+ * @author Pimgd
```
Pimvgd: is this okay

Innoxia: Yeah ^^

PR #41, opened 2017-08-26
fix buyback not clearing item reference

Pimvgd:
this stops the player from buying back the same item multiple times


Pimvgd:
tossing this on the "fix all inventory bugs" PR that will come around some time tomorrow probably

PR #42, opened 2017-08-26
Clean a bunch of random things

CognitiveMist:
Things I changed:
- List sorting uses comparator class
- Use addAll instead of for loops
- String comparison don't use ==
- Javadoc stuff
- Spelling
- Change some variables to local, that were only used in one spot
- Eliminated some dead/constant logic
- Any other cleanup that I could find

Things I ignored:
- Methods, etc. that are public but could be private
- Unused methods (getters/setters and more)

I limited this change to only a few files, partly because I want to make
sure Innoxia's OK with this (or if I shouldn't be quite so heavy-handed
in some ways), and partly because sleep.


CognitiveMist:
It just occurred to me I may have gotten a bit carried away... I guess this matches my idea of fun? :P

I will gladly continue on this sort of thing, if you like, though maybe in batches of only 1 or 2 files at a time.

Pimvgd:
I reviewed this and don't think it will break anything that wasn't already broken. Replacing != "" with ! .isEmpty() might break stuff but the alternative was printing null or crashing a couple lines earlier or later so I'd say merge it and we'll see if anything breaks in playtesting.

Phlarx:
I did some more reading, and it may be even better to use "".equals( ), which allows us to avoid an explicit null check. (It will also attempt reference comparison, if that happens to be the case; this is the only case checked before this change). Let me know if you'd like me to go that route.

CognitiveMist:
Oops that was a different account. Oh well.

Innoxia:
This a large amount of stuff for me to read through, and I need to get some rest this weekend. I'll read through and merge it early next week (Monday or Tuesday).

Thank you for doing all this CognitiveMist! ^^

CognitiveMist:
Ok, sounds good! Take your time, please. I wanted to find a way to show my support and help out, but money's not exactly abundant for me right now.

Thanks for all the love, time, and effort you've put into this!

src/com/base/game/Game.java
```diff
@@ -1642,10 +1638,7 @@ public boolean isStarted() {
 	}
 
 	public void unlockClothing(ClothingType clothing, Colour colour) {
-		if (unlockedClothes.get(clothing) == null)
-			unlockedClothes.put(clothing, new HashSet<>());
-
-		unlockedClothes.get(clothing).add(colour);
+		unlockedClothes.computeIfAbsent(clothing, k -> new HashSet<>()).add(colour);
```
Innoxia: I haven't used this function before, so it confused me for a moment. ^^

src/com/base/game/character/GameCharacter.java
```diff
@@ -1281,18 +1274,10 @@ public int getTotalTimesHadSex() {
 	
 	// Sex:
 	public void incrementSexCount(SexType sexType) {
-		if(sexCountMap.get(sexType)==null) {
-			sexCountMap.put(sexType, 1);
-		} else {
-			sexCountMap.put(sexType, sexCountMap.get(sexType) + 1);
-		}
+		sexCountMap.merge(sexType, 1, (a, b) -> a + b);
```
Innoxia: There are so many of these helpful methods that I never use. ^^

src/com/base/game/character/GameCharacter.java
```diff
@@ -1437,7 +1410,7 @@ public String incrementHealth(float increment) {
 					Combat.getOpponent().incrementMana(manaLoss);
 					setHealth(getHealth() + increment);
 					
-					return (UtilText.parse(this,
+					return (UtilText.parse(Combat.getOpponent(),
```
Innoxia: Nice catch ^^

src/com/base/game/character/body/OrificeUrethra.java
```diff
@@ -149,27 +149,27 @@ public String setCapacity(GameCharacter owner, float capacity) {
 			if (owner.isPlayer()) {
 				return "<p>"
 							+ "You let out a shocked gasp as you feel your urethra dilate and stretch out as its internal [style.boldGrow(capacity increases)].</br>"
-							+ "The transformation quickly pcockes, leaving you with [style.boldSex(" + UtilText.generateSingluarDeterminer(capacityDescriptor) + " " + capacityDescriptor + " urethra)]!"
```
Innoxia: Oh dear, that must have happened when I replaced all instances of "ass" with "cock"...

src/com/base/game/character/effects/PerkTree.java
```diff
@@ -39,10 +40,9 @@
 	static {
 		// First, order perk trees by their lowest level (first) perk:
 		List<PerkTree> perkTreesList = new ArrayList<>();
-		for (PerkTree pt : PerkTree.values())
-			perkTreesList.add(pt);
+		Collections.addAll(perkTreesList, PerkTree.values());
 
-		Collections.sort(perkTreesList, (a, b) -> a.getLinkedPerks()[0].getRequiredLevel().getLevel() - b.getLinkedPerks()[0].getRequiredLevel().getLevel());
+		perkTreesList.sort(Comparator.comparingInt(a -> a.getLinkedPerks()[0].getRequiredLevel().getLevel()));
```
Innoxia: Witchcraft!

src/com/base/game/dialogue/utils/OptionsDialogue.java
```diff
@@ -258,7 +258,7 @@ public String getHeaderContent(){
 					+ "<th></th>"
 					+ "</tr>");
 			
-			Collections.sort(Main.getSavedGames(), (e1, e2)->{return e1.lastModified()>e2.lastModified()?1:(e1.lastModified()==e2.lastModified()?0:-1);});
+			Main.getSavedGames().sort(Comparator.comparingLong(File::lastModified).reversed());
```
Innoxia: More witchcraft!

src/com/base/game/dialogue/utils/PhoneDialogue.java
```diff
@@ -1619,11 +1618,10 @@ public String getHeaderContent() {
 								journalSB.append("<td class='perkCell'></td>");
 
 						} else { // Arrow icon
-							if (i < 8) {
-								if (PerkTree.getArrowGrid()[i / 2][j - 1] == true) {
```
Innoxia: Forgive me ;_;

src/com/base/game/dialogue/utils/UtilText.java
```diff
@@ -3229,17 +3224,7 @@ public String parse(String command, String arguments, String target) {
...
-				return 0;
-				
-			if(e1.getRelatedBodyPart() == null)
-				return -1;
-			if(e2.getRelatedBodyPart() == null)
-				return 1;
-			return e1.getRelatedBodyPart().compareTo(e2.getRelatedBodyPart());
-			});
-		
+		commandsList.sort(Comparator.nullsLast(Comparator.comparing(ParserCommand::getRelatedBodyPart)));
```
Innoxia: So much witchcraft!

src/com/base/game/character/GameCharacter.java
```diff
@@ -2918,14 +2889,10 @@ public boolean isFeminine() {
 					isFeminine = true;
 					break;
 				case CLOTHING_FEMININE:
-					isFeminine = true;
-					if(getClothingAverageFemininity()<50)
-						isFeminine = false;
+					isFeminine = getClothingAverageFemininity() >= 50;
```
mirh: Ehrm.. This seems wrong I presume?

CognitiveMist: Thanks for your interest, but I'm not sure what you mean? I fairly certain those two do the same thing. If the femininity is 50 or above, it's true; if below, it's false.

Also, this changeset is 2+ years old; the code as seen here may have changed substantially since this was merged.

mirh: No it didn't. I came as far back as here exactly because blame brought me. 

And the part that I'm not understanding is why masculine should turn out positive when femineity is higher.

CognitiveMist: 0 is full masculine, 50 is the midpoint, and 100 is full feminine. All values are positive.

mirh: I see, thanks for replying. 
But then why is there that "greater than" sign for CLOTHING_MASCULINE?

DrZed:
`isFeminine = true;
					if(getClothingAverageFemininity()<50)
						isFeminine = false;
					isFeminine = getClothingAverageFemininity() >= 50;`
could be simplified to just the other bits are excessive and a waste of computational load, negligible but still a waste.
`isFeminine = getClothingAverageFemininity() >= 50;`

src/com/base/game/character/GameCharacter.java
DrZed: `isFeminine = true;
					if(getClothingAverageFemininity()<50)
						isFeminine = false;
					isFeminine = getClothingAverageFemininity() >= 50;`
could be simplified to just this. the other bits are excessive and a waste of computational load, negligible but still a waste.
`isFeminine = getClothingAverageFemininity() >= 50;`

CognitiveMist: mirh: I think you might be misinterpreting what CLOTHING_MASCULINE means. For convenience, I've copied the whole function here:
```
	public boolean isFeminine() {
		boolean isFeminine = body.isFeminine();
		
		if(Femininity.valueOf(getFemininityValue()) == Femininity.ANDROGYNOUS) {
			switch(Main.getProperties().androgynousIdentification){
				case FEMININE:
					isFeminine = true;
					break;
				case CLOTHING_FEMININE:
					isFeminine = getClothingAverageFemininity() >= 50;
					break;
				case CLOTHING_MASCULINE:
					isFeminine = getClothingAverageFemininity() > 50;
					break;
				case MASCULINE:
					isFeminine = false;
					break;
				default:
					break;
			}
		}
		return isFeminine;
	}
```
The values FEMININE, CLOTHING_FEMININE, CLOTHING_MASCULINE, and MASCULINE are determined by the gender settings in the game options. Specifically, this function determines whether an androgynous character should be described as 'feminine' and use feminine words like 'daughter'.

The two options that start with CLOTHING_ look at the femininity/masculinity of that character's clothes to make the decision. In both cases 0 to 49 will be masculine, and 51 to 100 will be feminine. The interesting part is if the value is exactly 50 -- their clothing is completely neutral (or they're not wearing any). In the case of CLOTHING_FEMININE, the value of 50 >= 50, is true, so the character is treated as feminine. In the case of CLOTHING_MASCULINE, the value of 50 > 50 is false, so the character is treated as masculine.

Hopefully that clears things up.

DrZed, I might be miunderstanding what you're trying to say, but I think you've misinterpreted what the show code snippet means. The red highlight shows what this code looked like before, and the green shows after. So the final result is just the green line, as seen in the function as its copied above.

DrZed: Yeah I was tired, and saw a post on something quite old and was slightly irritated due to various factors of the day. Tired due to 37 hours awake at that time, I've slept since then

DrZed: Yeah I was tired, and saw a post on something quite old and was slightly irritated due to various factors of the day. Tired due to 37 hours awake at that time, I've slept since then

mirh: Holy macaroni, now I see that. Sorry for the misunderstanding.

PR #47, opened 2017-08-28
Fix for ToeNail makeup

Rfpnj:


Rfpnj:
Changes have been folded into Moo2 for convenience and to make sure they did not get patched over.

PR #58, opened 2017-09-04
split code and resources, add gradle build tool

sweatwiring:
Added the Gradle wrapper to make builds easier ('gradlew build' in a command line window to create a runnable JAR with only the Java SDK).

Both Eclipse and IntelliJ support project workspace setup through importing a Gradle build file, so dev environment setup is simpler as well.

The Gradle convention is to separate source files and resources in two folders ('src/main/java' and 'src/main/resources') which is why there are a thousand file moves.


Rfpnj:
It seems to be a lot of changes, just to aid some people with building the game. Every file has been renamed.

sweatwiring:
It's possible to configure the source and resource directories, I've moved the source files back to src/ and moved the svgs to res/ to limit the number of nested directories.

There's some other advantages to using a build tool. I've added an existing launch4j plugin that'll generate self-contained executables with 'gradlew createExe'. (I think I saw some mention of launch4j as a goal.) It can be configured to also include the JRE if desired.

Gradle also supports easy dependency resolution for third party libraries as well as the modularization of projects. As the project grows in complexity, one or both might be needed.

Innoxia:
I'm sorry that it took me so long to respond to this PR - I've never used a build tool like this before, so I was (and still am) very hesitant to merge such a huge number of changes without understanding what it's doing. I recently merged #115, which (I think) does what you were trying to do with this PR?

Sorry again that I didn't get back to you for so long! :(

PR #60, opened 2017-09-05
Clean various issues

Pimvgd:


Innoxia:
I've gone through all of these changes and implemented them into my local version of the code. Thank you for this, and I'm so sorry it took me so long to do this! :3

PR #62, opened 2017-09-07
Some performance improvements

CognitiveMist:
Use Collections.addAll instead of iteration
Use StringBuilder instead of String += in loops
Etc


Innoxia:
Thank you! :3

PR #67, opened 2017-09-12
Changes Bovine Penis, from flared to tapered

Rfpnj:


Innoxia:
I've implemented this in my local code, thanks Rfpnj! ^^

PR #80, opened 2017-11-05
Util.intToString optimization and removal of EnchantmentDialogue.getRarityIdentifier

Briangulator:
Not a big contribution, but it's something.

When a number between 0 and 20 (inclusive) is input, the new code completes in 1/3 to 1/2 of the old code's time.
When a number greater than 20 is input, the new code completes in similar time as the old code.
The new code's class file is 103 bytes smaller than the old code's file.


Briangulator:
I poked around some more and found a static method that doesn't seem necessary, so let's get rid of it.

Innoxia:
Thank you Briangulator! ^^

It looks like my comment "// This seems stupid... ;_;" was right after all... I don't know what possessed me to use a switch statement there. ^^

PR #82, opened 2017-12-05
Adjust prostitution odds/prices

tukaima:
This code will make it so the odds of any random NPC becoming a prostitute scales based on their body's features, and also includes code to adjust price based on the prostitute's features. Code is probably kind of bad, but you can modify the odds and price separately.

Currently the odds are set so that feminine prostitutes show up at the same rate (50% for a female with average breasts.) Horses are heavily weighted towards becoming prostitutes since they have big dicks and big breasts. Male characters have around a 25% chance to be prostitutes, but this scales based on the length of their dick.

### DIFF
--- a/src/com/lilithsthrone/game/character/CharacterUtils.java
+++ b/src/com/lilithsthrone/game/character/CharacterUtils.java
@@ -884,7 +884,7 @@ public class CharacterUtils {
 						.horn(new Horn((stage.isHornFurry()?startingBodyType.getRandomHornType(false):HornType.NONE), (startingGender.isFeminine() ? startingBodyType.getFemaleHornLength() : startingBodyType.getMaleHornLength())))
 						.antenna(new Antenna(stage.isAntennaFurry()?startingBodyType.getAntennaType():AntennaType.NONE))
 						.tail(new Tail(stage.isTailFurry()?startingBodyType.getTailType():TailType.NONE))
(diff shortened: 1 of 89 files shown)


Innoxia:
Thank you! I'll try and get this merged over the weekend! :3

src/com/lilithsthrone/game/character/CharacterUtils.java
```diff
@@ -928,7 +928,7 @@ public static void randomiseBody(GameCharacter character) {
 		}
 		
 		//Ass:
-		if(character.hasFetish(Fetish.FETISH_ANAL_RECEIVING) || character.getHistory()==History.PROSTITUTE) {
```
Innoxia: I added in the prostitute background check, as I assumed that the NPC would have lost their anal virginity due to their line of work. The same is for the two cases below (prostitutes would have lost their vaginal virginity, and wouldn't have a pure virgin fetish).

Innoxia: I probably should have read through all of your changes before commenting... It makes sense now... Sorry! x_x

Innoxia:
Thank you so much for the work you've done tukaima! I made a few changes on my end (mostly moving your code into new methods in the CharacterUtils class), and have implemented it into my code! ^^

Sorry that I didn't directly pull it from here on github, but I've credited you both in the patch notes and as an author in the CharacterUtils class (where I relocated your new code to). I'll push my local branch to github now. :3

Thank you! <3

PR #89, opened 2017-12-23
Fix demon wings incorrectly described as allowing flight

tukaima:
Wings which do not allow flight (currently, demon wings and NONE) should never be described as allowing flight if they are of average size or larger.

### DIFF
--- a/src/com/lilithsthrone/game/character/body/Body.java
+++ b/src/com/lilithsthrone/game/character/body/Body.java
@@ -2549,7 +2549,7 @@ public class Body implements Serializable, XMLSaving {
 				default:
 					break;
 			}
-			if(wing.getType()!=WingType.NONE) {
+			if(wing.getType().allowsFlight()) {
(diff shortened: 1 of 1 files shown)


Rfpnj:
This commit does not consider the size of the wings, for judging whether they are flight worthy. A better commit would be to add the override to allow demon wings to fly until the demon revamp is done.

src/com/lilithsthrone/game/character/body/Body.java
```diff
@@ -2549,7 +2549,7 @@ public String getDescription(GameCharacter owner) {
 				default:
 					break;
 			}
-			if(wing.getType()!=WingType.NONE) {
+			if(wing.getType().allowsFlight()) {
 				if(wing.getSizeValue() >= WingSize.TWO_AVERAGE.getValue()) {
```
tukaima: It already does exactly that.

tukaima:
The base code does exactly that, though.
`if(wing.getSizeValue() >= WingSize.TWO_AVERAGE.getValue())`

PR #91, opened 2017-12-23
Removed an "unneeded" check (domionion dialogue bugfix)

Lucasvdlaan:
There was a check in the intToString that checked if the number was
Bigger than 0, yet for the "times" bug, you'd need the zero to be done
as well.

Also on line 357 there was getLostPlayerCount instead of
getFoughtPlayerCount


Innoxia:
Thank you! :3

PR #93, opened 2017-12-26
Dom npc's can ignore player pull out requests

throwaway1003:
The isBaseRequirementsMet method of the
 PARTNER_GENERIC_ORGASM_CUM_INSIDE sexAction is modified to allow
 dominant npc's to ignore pull-out requests. If the penetration is
 vaginal, npc's with seeder and impregnation fetishes will always
 ignore the player. Otherwise the chance is 50%. This value can be
 easily changed by modifying the value on line 1131.
 
There appears to be some non-functional code in SexManagerDefault
 intended to do this. This change should serve the same purpose.


throwaway1003:
Sorry, new to git/Github. I thought I was only working in a fork. Feel free to use it if you like the change. Otherwise, no offense taken if you close/reject without comment.

PR #94, opened 2017-12-27
Implement dominance and recessiveness in genes.

tukaima:
Wrote most of the changes down already in my first commit, so I'll just copy those here.

> Races now have a property that adjusts just how dominant or recessive
their genes are. This is added or subtracted to the base 50% chance for
offspring to resemble the mother.
> Demons add 100% to the chance for offspring to resemble them, making it
150% in favor of them.
> Slimes add 50% to the chance for offspring to resemble them, making it
100% in favor of them.
> Angels subtract 50% from the chance for offspring to resemble them,
making it 100% in favor of their partner.

> In addition to the race-based numbers, characters can also adjust these
odds by spawning in with a fetish. Currently, these fetishes adjust odds
by a flat 40%. A Demon with recessive genes adds 60% chance for
offspring to resemble them, and this lowers to 20% if their partner also
has dominant genes, so it's possible (albeit difficult) to have demons
create non-demon children. This happens with a 30% chance with the
mentioned combination of fetishes.
> Need new art for the new fetishes, they are presently using those for
Dominant and Submissive.


Rfpnj:
I like the idea of the fetishes, however, I think they would be better served as perks then fetishes.

tukaima:
There is some basis for having similar features as perks (particularly, Barren), but I opted for fetishes because they can easily be granted to NPCs, and they're also infinitely more accessible to players, since they aren't locked by level and use an easily renewable resource.

Could also assign "dominance"/"recessiveness" to the penis/vagina and allow it to be modified with TF potions, but since furry races can have humanoid genitals rather than furry genitals (based on furry preferences), most races would end up using human values at lower furry preferences, when they should probably be consistent regardless of furry preference.

Rfpnj:
I can see your point about ease of use, however, the fact being a perk would make it harder to impose on NPCs really should not be an issue and makes it harder to change. Perks tend towards permanence while fetishes are temporary.

tukaima:
This PR no longer uses fetishes, as described in the first comment. Instead, "autosomal dominance" can now be adjusted with transformation potions. This property is under the "core", and ranges from -500 to 500.
Races can also have autosomal dominance defined separately for males and females. Angels (-100), Slimes (500), and Demons (100) still have non-zero values, primarily for testing. Given that the lore for Angels and Demons says that they breed normally, feel free to change these back to 0.

Rfpnj:
The changes make more sense to the world and its lore.

Innoxia:
Sorry that it's taken me so long to get to this! I'll be improving the genetics system very soon, and will merge this PR when I get to it! :3

tukaima:
Hi. I updated this thing. Force push. Bug me if you need something from the older version.

Save compatibility is fucked as of this latest commit. Saves which specifically lack genesRate values will not load. This is probably tied to loadFromXML under the Body class.

CognitiveMist:
@tukaima 

This PR is out of date with most of the code it is based upon. Do you still believe it is useful? Do you think it might be worth updating?

PR #96, opened 2018-01-02
Support for zero nipples

tukaima:
Fixing nipple-related stuff to allow characters and races with 0 nipples.


tukaima:
Closing this because all of this code is just going to be in the Gargoyles PR.

PR #97, opened 2018-01-04
Implementing RacialSelector, Subspecies Frequency preference options, Gargoyle leftovers, support for no nipples

tukaima:
RacialSelectors are hashmaps that assign odds of appearing to each variation of any given race. These will be used for cases where we're interested in their general body type rather that specific ones, like when generating a random NPC. As an example, Fox-morphs, "kitsune", and "pipefoxes" could all be part of a fox RacialSelector even if they would be separate RacialBodies. If a method needs a RacialBody, you can provide a random variation from the given selector by using RacialBody.getBodyFromSelector(RacialSelector.SELECTOR_NAME).

This commit also includes leftovers from my unfinished work on Gargoyles. At lesser stages, gargoyles will actually spawn in without any gargoyle parts whatsoever (due to the way their RacialBody is designed) and they also don't correctly use furry preferences tied to other races.

This also includes all the code in #96.


tukaima:
As of the last commit, Gargoyles are in a semi-usable state and should respect the furry preferences for other races.

tukaima:
Being that subspecies has been implemented (albeit, in a different fashion) there really isn't much point to keeping this open right now. Closing.

PR #100, opened 2018-01-06
Fix clothing tooltips in the encyclopedia

CognitiveMist:
- dyeClothing was passed to the tooltip as null
- Secondary/tertiary colors were orangey by default?
- PhoneDialogue was using old id still, but no clothes were in that dict
- Also, in Colour, move masculinePlus name from MASCULINE to MASCULINE_PLUS


Innoxia:
The clothing tooltips were fixed in 0.1.97, but I have applied your "move masculinePlus name from MASCULINE to MASCULINE_PLUS" fix into my local branch. Thank you! :3

PR #102, opened 2018-01-12
Removed Arthur's Package and Hypno Watch from object that an be finded in alleyways

mephistophele:
Removed ARTHURS_PACKAGE and ORIENTATION_HYPNO_WATCH from commonItems


CognitiveMist:
A pedantic thing, but please try to match whitespace styles with existing code. Innoxia used tabs here, so edits should too.

Innoxia:
Thank you very much for the fix, but I changed the way alleyway spawns are handled in 0.1.97, which has solved the problem. ^^

PR #103, opened 2018-01-15
Linux font compatibility

PyrophoricPlumage:
This branch adds Carlito as an alternative font for Calibri in the CSS. This is necessary to display the interface correctly when the game is run natively on Linux. (There's no Calibri and the text becomes too large for the buttons.) Carlito is supposed to be an open-source alternative to Calibri with the same font metrics, so no sizes change.

Note: The .jar from the website has been tested and works correctly (except for the font issue) on Ubuntu 17.10, so you may want to add that info on the website. (It used to crash in an earlier version, but I don't know which of the many package updates has fixed the issue.) The following packages and their dependencies have to be installed for the fixed version: `openjfx fonts-crosextra-carlito`.

### DIFF
--- a/src/com/lilithsthrone/res/css/webViewAttributes_stylesheet.css
+++ b/src/com/lilithsthrone/res/css/webViewAttributes_stylesheet.css
@@ -1,6 +1,6 @@
 html {
 	background: #1e1e20;
-    font-family:Calibri;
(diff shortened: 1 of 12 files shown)


Innoxia:
Thank you! :3

PR #106, opened 2018-01-16
Changing striped clothing items to use CLOTHING instead of BASE colors.

tukaima:
Striped leggings, panties, and bra incorrectly use BASE_BLUE_LIGHT and BASE_PINK_LIGHT, causing the color viewer for these clothing items to erroneously display both the BASE and CLOTHING variations of these colors. This fixes that issue.


Innoxia:
Thank you! :3

PR #111, opened 2018-02-03
Update lilithsThroneBuildTutorial.md

picobyte:
On some Linux distributions eclipse may be configured with openjdk by default.
Dditional small changes to make instructions more clear.


Innoxia:
Thank you! :3

PR #112, opened 2018-02-07
Couple of sex fixes

MissyDirection:
Currently, if a player is getting tail-pegged while on top during chair sex, they have no ability to stop it, which in turn means no changing positions. This fixes that.

Tested with Kate.

Also correct pulling the wrong orifice/penetration description on initial penetration.


Innoxia:
Thank you very much for this! <3

PR #115, opened 2018-02-12
Add Maven build system

pinnae:
Was tryna get the game building on my computer, but since I don't use Eclipse I used Maven
Managed to get it working without modifying the folder structure like #58 did with Gradle

I think Eclipse has support for Maven, but if you wanna build from the command line you can just do
`mvn package` and then `java -jar target/game-1.jar` to run


Innoxia:
Thank you for this! I'm sure it will help a lot of people! ^^

PR #117, opened 2018-02-24
Encoding compatibility for special characters

CognitiveMist:
This a big, fast pass to make the following changes:
 - Smart qoutes -> simple quotes
 - Accented characters/symbols -> html/xml entities (e.g. &eacute;)
 - En/em dashes -> simple dashes
 - Ashley's height uses the standard npc.height... parser

Since this was a bulk job, I also converted all the .txt files too; this might be useful
if copy/pasting from those files, or it might be overkill.

For reference, the command I used to locate the above potential issues is:
	`grep --color -n -P -r "[^\x00-\x7e]" --exclude="*.png" res/ src/`
...which on my system now returns only bugs.txt (it's the BOM, this is fine), and the
two currency symbols I left as comments.


Innoxia:
Thank you for this! I've applied the fixes on my end using the work you've done. ^^

PR #118, opened 2018-03-01
Orientation Bug

A-Master-of-Puppets:
This is a fix for the bug where the attacker and defender are the wrong way around when comparing orientation to femininity.


Innoxia:
Thank you! Would you like me to add you into the patch notes (as a contributor) as "A Master of Puppets"?

A-Master-of-Puppets:
I would prefer just "Master of Puppets", if that's OK. The "Master of Puppets" handle was already taken on Github.

PR #119, opened 2018-03-01
Tail Actions

A-Master-of-Puppets:
This moves actions involving tails from the specific position action lists into the generic action list, making them available in all generic positions. I haven't touched any positions that don't inherit the generic action lists.


Innoxia:
I'm not sure if tail actions should be in the generic action list. This would make it so tail actions are available in every position... I guess it depends on how long the tail is, but I imagine that it would be difficult to have tail-mouth interaction when the person with the tail is behind in doggy-style, for example.

A-Master-of-Puppets:
I don't see why not; if I'm behind someone in doggy, my tail is coming from pretty close to where theirs is, and they can suck their own just fine. It _does_ depend on how long the tail is (and if it can stretch; demons are magic), but I got the impression they were pretty long. I could add the actions on a position by position basis, it just seemed simpler to make them general.

Innoxia:
Well, I meant for the 'playerGenerics'/'partnerGenerics' to only contain actions that will be available in any conceivable position. I don't know if tail actions will *always* be available in every single position...

PR #120, opened 2018-03-01
Cum Targets

A-Master-of-Puppets:
This adds body parts to the lists of targets that a character with a dick can cum on for various positions. It doesn't allow characters to cum on themselves; that doesn't appear to be possible without engine work.


Innoxia:
I like the cum target additions, but not the removals (from missionary and cowgirl).

A-Master-of-Puppets:
Whoops! Those changes weren't supposed to be there! That's just something I did for myself until we get to control where our partner cums, because I was tired of them cumming on me.

Innoxia:
Ah I see. ^^
Thank you for this! :3

(P.S. Are you on the LT discord? If so, PM me on there and I can give you the contributor role. ^^)

PR #147, opened 2018-03-04
Fix for clothing colours bug

A-Master-of-Puppets:
The game was only checking _default_ colours on import, not available dye colours. So, if you dyed your clothes a colour they couldn't generate with, the colour wouldn't load.


Innoxia:
Thank you! <3

Issue #126, opened 2018-03-03
Clothing effects Enchantment BUG


riwigica:
After reaching the maximum height of the enchantment it will revert 122cm for example (Major Boost limit 270cm):

```
(...) You are now 269cm tall. (...) 
(...) You are now 270cm tall. (...)
(...) You are now 148cm tall. (...)
(...) You are now 149cm tall. (...)
```

and the other enchantments afther reaching the maximum will spamm  the screen for example:

```
 "Your body's muscles don't change...".
 "Your body's muscles don't change...".
 "Your body's muscles don't change...".
 "Your body's muscles don't change...".
```

I think you may have forgoten a "=" sign in:

```javascript
//liliths-throne-public/src/com/lilithsthrone/game/inventory/item/ItemEffectType.java 
private static boolean isSetToLimit(int increment, float currentValue, int limit) {
//(...) line 3194
			if(currentValue <= limit) {
//(...) line 3198
			if(currentValue >= limit) {
//(...)

Innoxia:
Thank you for the detailed bug report! <3

PR #153, opened 2018-03-04
Fix incorrect inequalities for set to limit check

A-Master-of-Puppets:
Fix from #126 implemented.


Innoxia:
Thanks for the fix! ^^

PR #155, opened 2018-03-05
Enchantment effect translate for old saves

A-Master-of-Puppets:
This just adds a check on importing ItemEffects to see if they refer to strength, fitness, or intelligence, and if so translate that to physique or arcane as appropriate. With this, players with saves from before the change should hopefully keep any potions they had made.


Innoxia:
Thank you! ^^

PR #156, opened 2018-03-05
Fix for export error with nulls in booksRead

A-Master-of-Puppets:
This simply skips over any nulls that might get into booksRead, stopping them from blocking export. It doesn't address why nulls might be getting into booksRead to begin with, just stops them from doing the damage if they do.


Innoxia:
I'm not sure how nulls are getting into that Set... :s 

Thank you for the fix! :3

Issue #159, opened 2018-03-05
Blog/Github (browser launch) buttons hangs Lilith


picobyte:
The blog button in the far right in file menu (with save, load etc.) hangs Lilith for me. When clicked, no webpage is opened and the game does no longer respond, effectively hanging the game.
I think the reason may be that lilith has no permissions to start the browser in linux, that is no reason to hang though IMHO.

Platform arch linux x86_64, jdk1.8.0_161/bin/java
running lilith git, latest commit: e685d1d

This button is at src/com/lilithsthrone/game/dialogue/utils/OptionsDialogue.java 183

The same problem occurs when pressing the github button.

picobyte:
Problem was also discussed here:

https://stackoverflow.com/questions/27879854/desktop-getdesktop-browse-hangs

picobyte:
added pull request https://github.com/Innoxia/liliths-throne-public/pull/160 to fix this

Innoxia:
Thank you for the report and fix! ^^

PR #160, opened 2018-03-06
fix opening an url

picobyte:
testing Desktop.isDesktopSupported() or
Desktop.getDesktop().isSupported(Desktop.Action.BROWSE)
did not work on my platform. So just resorted to the other way around
approach.

Signed-off-by: picobyte <PiC.o0.o.BytE@gmail.com>


picobyte:
To resolve issue #159
Tested in linux. Does this work on other platforms as well?

Innoxia:
I can confirm that this works on my system as well (Windows 7). I'll merge it now - thank you for the report and fix! ^^

PR #162, opened 2018-03-06
Fix for fatal bug on cultist sealed scene

A-Master-of-Puppets:
The cultist sealed scene had the wrong starting positions, and Sex.updateAvailableActions() does _not_ handle this gracefully. I've corrected the bad positions, but sex positions should really have default positions to fall back on in case this happens again.


Innoxia:
Thank you! :3

PR #163, opened 2018-03-06
Boxers and briefs displace update

A-Master-of-Puppets:
Changes boxers to use 'unbutton' and 'pull down' for displacement, and crotchless briefs to use 'pull down'. Also adds concealment slots to them; if I've done this right, I might have a go at adding them to other items.


Innoxia:
Thank you for this! There are some underlying bugs in the way clothing displacement/removal is being calculated (which appears to surface whenever there are multiple displacement types covering the same areas), so I'll merge this as soon as I've got those sorted out. :3

Innoxia:
I've implemented this in my local code, and it will make into the next commit. I'll credit you in the patch notes, thank you for this! :3

PR #164, opened 2018-03-06
Prologue human race handles NPCs now

A-Master-of-Puppets:
Previously, Alexandria's "human" status effect said "You're a human, just like every other person in this world."


Innoxia:
Thank you! ^^

Issue #128, opened 2018-03-04
Dirty clothing icon missing in inventory


Innoxia:
Dirty clothing has an indicator icon overlaid while it’s equipped, but not when it the inventory.

picobyte:
an attempt to address this: pull request #165
In buyback modus the cum splatters are still not visible. Does that also need to be fixed?
Also, shouldn't price decrease for cum-stained clothing?

Innoxia:
Fixed in #165.

I don't think it matters too much if it doesn't display in the buyback menu - I'm going to be improving that part of the UI at some point. ^^

I will add a small price decrease for cum-stained clothing, good idea!

PR #165, opened 2018-03-06
Fix issue Dirty clothing icon missing in inventory #128

picobyte:
Now the cum splatter is shown even when not equipped. Did some testing, seems to work

Signed-off-by: picobyte <PiC.o0.o.BytE@gmail.com>


Innoxia:
Thank you very much! I'm sorry that you had to suffer reading through my horrible code in the `appendDivsForItemsToInventory()` method! ^^

PR #166, opened 2018-03-06
Add various modifiers to makeup coverings

A-Master-of-Puppets:
Adds a number of modifiers to some makeup types, and enables their use in the description pages. I'm not an expert on makeup, so there might be more that could be added.


Innoxia:
This is cool - thank you! :3

PR #167, opened 2018-03-07
Couple typo fixes

WoefulWombat:
I hope I'm doing this right, it at least looks like it. Takes a while to get into Github.
Not much to say. The second "kimonos" is maybe debatable though, and I hope getNameSingular is really the correct way to get the singular name.


Innoxia:
Thank you for the typo fixes! Yeah, you've done it all correctly! :3

(I love your avatar by the way - flying witch is great. ^^)

PR #168, opened 2018-03-07
Fix for importing potions

A-Master-of-Puppets:
It is Attribute.MAJOR_ARCANE, but ItemEffectType.MAJOR_ARCANE doesn't exist. Potions still use ATTRIBUTE_ARCANE for the itemEffectType field.


Innoxia:
Oh, I'm so sorry, you already fixed this once before didn't you? ;_;
I didn't realise what I was doing, and broke it again... x_x

Thank you for the fix!

PR #169, opened 2018-03-08
Small descriptor change

Lightcanadian:
i was confuse in the enchanting menu when the fetish had almost no description so i added a description in the enchanting menu and change the description of the breast fetish that was too vague


Innoxia:
Thank you for this! I've implemented this on my end, so, while I won't pull this request (as I've just implemented it quickly myself), I'll credit you for it in the patch notes! :3

PR #172, opened 2018-03-10
 bladder

picobyte:
This adds a framework for bladder control, a few icons, a bladderLevel attribute, statusEffects. An attempt is made to get the tooltip, but this does not show yet - ~~not sure why~~ **Edit:** fixed in local branch.
The code seems to run.


picobyte:
Althoug hit should be safe to merge, maybe not do that just yet, lets have some discussion first.
I was thinking about adding some toilets in several sites. That the bladder gradually increases in time/number of turns.

picobyte:
Instead of the bar, maybe we could make an icon instead, if the character needs to go. 

A possiblity is to make a TractStomach and TractBladder and maybe a TractIntestines that processes what passes through mouth (class), tract classes a bit similar to the character/body/Orifice* classes. and fluids finally end up in the bladder (after some time in the body).
Something similar could be done to process food in the TractIntestines - not my kink in particular, but anatomically correct.

I could imagine a body overview - one for a male, another for female and futinari - with several tabs:
* clothing coverage (no attempt to draw all clothings, just indicate what bodypart(s) are covered).
* orifice and tract (organ) status: shown is an anatomical overview of the body. Hoover delivers status per organ. All the pervs probably would like to be able to put something in an orifice and see it here.

picobyte:
I added a diureticum that acts upon a bladder via player.addPotionEffect(Attribute.BLADDER, 10); how can this be undone? player.setPotionAttribute(Attribute.BLADDER, 0f); doesn't seem to do it - statusEffects won't go away.

picobyte:
To test, go to toilet (Lilaya 2nd floor, south-west) and take 40+ diureticum (to speed up urine building - without this it will take a long time to rise). After walking around a bit, you'll get the bladder icon, it increases a few times before you pee your pants (unless you enter the toilet cell).
bladder status can be seen in phone->body stats

MissyDirection:
I would suggest that it's very important for this particular content to be able to be disabled through the game settings, and probably not enabled by default.

picobyte:
So in case you think this is about watersports, it is not, currently. This is about body function. Or what exactly is your concern?

A-Master-of-Puppets:
I can't see why anyone would want this _apart_ from watersports. It's completely unnecessary over-simulation, except for people who are into that kind of thing. The game is not going to be _improved_ by making you visit a bathroom a couple times a day, that's just going to be tedious.

By the way, the bedrooms in Lilaya's house have en-suite bathrooms, so adding a new room for that doesn't seem necessary either.

picobyte:
well I guess this is not my kind of game then. Would have been nice if it was stated it earlier. All this work for nothing.

A-Master-of-Puppets:
Don't get me wrong, I'm not saying that it shouldn't be included, I'm just saying that I really don't see many people toggling this on except for those who are into watersports. That's just my opinion, and I might be wrong. I see making the player visit the bathroom as more of a nuisance rather than actual content, something you have to do simply to avoid a debuff, and not for it's own sake. I don't see how going to the bathroom by itself can be anything except mundane. You could perhaps change this with some "fucking in the bathroom" scenes, or maybe gloryholes. Those would be goings on in the bathroom that would actually be _interesting._

picobyte:
Thanks for keeping the discussion open then.

As I said I'm not sure whether I am on the same track, but what I find exciting is when a character is gradually drawn into shameful situations. Situations that he/she would normally avoid.

Society imposes certain rules of normalcy, but everyone also has their own semi-subconscious desires that suffer a bit from these restrictions we impose upon ourselves. In a game we can have a character, representing ourselves or our desired counterpart, live a life wherein these rules of normalcy fall apart.

The protagonist is forced to make choices (limited choices for the player) and hence obtain the fetishes the player desires the protagonist to have. This I would rather have, than being able to simply select fetishes.

Maybe this is too far astray from your intent, and incompatible.

Innoxia:
I'm it's taken me a while to respond to this - I meant to comment on this earlier, but I got caught up with other things. ^^

I am definitely going to add watersport content to the game (behind an optional toggle), and this PR looks really interesting! As it's so large (5k additions at the time of this comment), it might take me a little while to look through it all, but I will merge it once I can check that it all fits into the game's setting! :3 

Thank you for taking the time to make this, and I'll try to read through it all over the weekend! (Please do feel free to @innoxia me at any time to remind me of this, or to ask any questions! ^^)

Rfpnj:
@picobyte Please update as people are asking for this and if Inno does not need to work as hard to add it, the faster it will get done. She has not forgotten about it.

picobyte:
Due to the delay there were several conflicts to resolve. I had to change some things to at least make it compileable. Something changed and now I cannot enable bladder or hunger/thirst in the character preferences screen. This makes testing impossible.

CognitiveMist:
@picobyte 

picobyte:
I'm not going to fix the merge conflicts.

Best,
R

Innoxia:
I'm sorry that I never got around to merging this. Thank you for the work you've done regardless, and I'll try and use some of your contributions when I get around to adding this content into the base game!

PR #181, opened 2018-03-10
Add coverage to clothing for slots with no sex actions

A-Master-of-Puppets:
Items of clothing are tagged for coverage for things like genitals or nipples, which need to be blocked for sex actions, but they have no coverage for things with no sex actions like stomach, back, or hair.

This adds such slots, which should fix the bug where you cum on a character's back or stomach and it dirties the body _through_ the clothes without touching the clothes.

### DIFF
--- a/src/com/lilithsthrone/game/inventory/clothing/ClothingType.java
+++ b/src/com/lilithsthrone/game/inventory/clothing/ClothingType.java
@@ -742,7 +742,15 @@ public class ClothingType {
 			null,
 			"head_cap",
 			null,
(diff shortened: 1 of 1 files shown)


Innoxia:
Thank you for this! I'll merge this tomorrow and test it thoroughly! ^^

Innoxia:
Sorry it took me a little while! ^^

I tested it, and it all seems to work fine. The reason that I was a little hesitant to merge at first is because of the underlying bugs in the clothing displacement system (the same concerns I have with #163). This is nothing wrong with your pull requests, but something that I need to sort out in the underlying code. ^^

PR #193, opened 2018-03-11
Bring partner 'resisting' action in line with player 'resisting' action

A-Master-of-Puppets:
This updates the PARTNER_RESIST action to recognise the NON_CON_SUB fetish. This should prevent NPCs with that fetish from lowering their arousal and therefore making it quite hard to get them to cum. I still feel that these actions probably shouldn't affect arousal _at all_ any more since that's what the new lust stat is for, but for now I've just made the two actions equivalent.


Innoxia:
Thank you! Yeah, I'll make an issue about that. ^^

PR #195, opened 2018-03-11
Punctuation, mainly commas

WoefulWombat:
I noticed some inconsistencies. As I learned it, vocatives need commas, even in short cases such as "Hello, mistress." or "Good evening, Lilaya.". I've let the "Hello" ones stand as is for now, since I wanted to ask you beforehand. Consistency is most important, so do you write "Hello, world." or "Hello world."? Or would you even use both, depending on the intended intonation?

And another one I left untouched: Narration preceding dialogue in the same paragraph, like:

_After a minute or two, the cat-girl's breathing has slowed to a regular rate, and she opens her eyes and smiles, "Sorry about that...."_

You use a comma to end the narration, but I learned to use a fullstop, unless the narration includes a dialogue tag or is interrupting a line of dialogue. Again, consistency is more important than guidelines, so what do you prefer?


Innoxia:
Thank you for this, WoefulWombat! I'll try to use the correct application of the vocative comma in my future writing! ^^

I'm not sure about the dialogue comma... I'll look into it! :3

PR #208, opened 2018-03-12
Add cum targets for characters cumming on themselves

A-Master-of-Puppets:
This adds new cum targets for a character in a reclining position (the bottom in missionary or cowgirl, for example) to cum on themselves rather than on their partner or the floor.

PR #209, opened 2018-03-12
Refactor GenericOrgasms.isGenericPartnerCumTargetRequirementsMet

A-Master-of-Puppets:
This refactors the GenericOrgasms.isGenericPartnerCumTargetRequirementsMet method, hopefully making clearer what it's doing, but more importantly removing the requirement that the partner be penetrating the player. This should fix the bug where, if the NPC is not penetrating the player, they have no cum actions and fall back to the PARTNER_ORGASM_GENERIC action, which always uses the first available cum target (rather than a random one), and also doesn't apply the dirty effect properly to either the player, or themselves if using my new cum targets from #208.


Innoxia:
Thank you! ^^

PR #210, opened 2018-03-12
Fix for incorrect lubrication on PC anal tease

A-Master-of-Puppets:
This is just a simple fix for the PC teasing an NPC's cock with their anus.


Innoxia:
Thank you! :3

PR #211, opened 2018-03-12
Add error logging

A-Master-of-Puppets:
This writes errors to a file rather than STDERR. This means anyone should be able to provide error logs when reporting bugs.

### DIFF
--- a/src/com/lilithsthrone/main/Main.java
+++ b/src/com/lilithsthrone/main/Main.java
@@ -1,6 +1,8 @@
 package com.lilithsthrone.main;
 
 import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.PrintStream;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Comparator;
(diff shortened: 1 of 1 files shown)


Innoxia:
I've slightly changed this to write to 'data/error.log' instead of just 'error.log'.

I also had to add '@SuppressWarnings("resource")' to the PrintStream. I hope this isn't a bad thing to do. :s

I'll push my changes (with the todo list crediting you)  immediately after closing this. Thank you! :3

A-Master-of-Puppets:
It's generally better to have error logs as high up as possible, so people don't have to go looking for them. I would expect the data folder to be for things that the game needs to read, not something write-only like logs.
Also, I'm not sure why you need to add a warning suppression; it didn't give any warnings for me. That's kind of weird, since all you've changed is the path and putting the stream in a variable.

Innoxia:
It was giving me `Potential resource leak: 'stream' may not be closed`.

Doing it this way creates the error.log file instantly (as soon as the game is run), even if there's no error, so I was concerned that people might get worried seeing the "error.log" file appearing, even though there's no error. That's why I put it in the data folder, so people wouldn't notice it and get worried. ^^

A-Master-of-Puppets:
Oh, that shouldn't be a problem then. It's just warning you that you open a file handle but don't close it. That would be bad if you did it while saving the game, say, but you don't _want_ to close the error log handle while the game is running. Still weird that eclipse didn't warn me about that, though.

PR #216, opened 2018-03-12
Fix for spell cost reduction giving negative costs

A-Master-of-Puppets:
This is my idea to stop very high spell efficiency from _gaining_ aura from casting spells. Rather than a linear function, it uses an exponential function, so instead of spells being free at 100 efficiency, they cost 33%, and rather than costing _minus_ 100% at 200 efficiency, they cost 11% of nominal. Theoretically, the cost should never reach zero, although of course with rounding this will happen eventually (a cost 100 spell would reach zero at 692 efficiency, for example).
This also causes costs to _increase_ exponentially as efficiency goes negative, of course. If this isn't desired, a simple if(efficiency < 0) check could restore linear behaviour for that case.

### DIFF
--- a/src/com/lilithsthrone/game/combat/Spell.java
+++ b/src/com/lilithsthrone/game/combat/Spell.java
@@ -692,7 +692,7 @@ public enum Spell {
 	public float getModifiedCost(GameCharacter caster) {
 		float calculatedCost = spellCost;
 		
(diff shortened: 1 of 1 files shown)


Innoxia:
Thank you for bringing my attention to this!

I've applied the formula I'm using for all value drop-offs to spell efficiency, so instead of your change, I've made it:
`calculatedCost *= ((100 - Util.getModifiedDropoffValue(caster.getAttributeValue(Attribute.SPELL_COST_MODIFIER), 100)) / 100f);`

PR #222, opened 2018-03-15
Body hair/fur/feathers pattern fix

A-Master-of-Puppets:
Body hair/fur/feathers should use the patterns for hair rather than skin - you don't get freckled fur or feathers. I'm not so sure about scales, but I think they probably shouldn't use the skin patterns either - I can change them easily if wanted.


Innoxia:
Good point! Thank you! :3
I guess scales shouldn't have the freckles option... I'll change that now. ^^

PR #225, opened 2018-03-16
Update Angry Harpy Quest to reflect new pricing scheme

Rfpnj:
Increases Flame award from Harpy quest to 5000 flames, which is now 1/3 of the cost for Scarlett, instead of 1500 (1/10 the cost). Where you do get a good deal of essences during the quest, with the increased need for them for use with enchanting it balances out a bit better.


Innoxia:
Thank you! :3

PR #226, opened 2018-03-16
Change what causes the "Frustrated" status effect

A-Master-of-Puppets:
Change what triggers the "Frustrated" status effect from "did not cum" to "ended sex just before orgasming". Also removed the requirement to not be in resisting pace; even if unwilling, being taken right to the edge and then stopped is going to be frustrating.


Innoxia:
Good idea, thank you for this!

PR #228, opened 2018-03-16
Add colours to characters

A-Master-of-Puppets:
Gives all GameCharacters their own colours. Things I've done:
 - Add screen for player to pick their colour during the prologue and in the phone
 - Use character colour for speech
 - Random NPCs take their colours from their clothing
 - Cultists take their colours from their skin colour
 - Works with saving and loading

Things I have _not_ done:
 - Given every static NPC their own colour. I've done some, but most I'm not sure what colour they should have. I'm not sure about all of the ones I _have_ done, to be honest.


WoefulWombat:
Letting the player choose their speech colour seems like a good idea, but how do NPCs now get theirs, exactly, from the colours of their clothing or the femininity/masculinity of the apparel? 
I quite like how it's handled now, and don't some unique NPCs already have their own colour? I remember Lilaya and Alexa being special there.

A-Master-of-Puppets:
The player colour is the most important part, yes. Although I found it kind of boring that all NPCs had the same five colours (which is really only five shades of _three_ colours), so I gave them their own.
As to what colours people get, it depends. With most random NPCs, the colour is primary clothing colour. Cultists, they don't use the regular clothing generator, so their colour is their skin colour. With unique NPCs, it will be the same custom speech colour if they have one (Lilaya, the Harpies, Zaranix's maids, etc.), or else I tried to pick some that seemed fitting. Brax is white for his fur, Rose is red for her hair, etc., although I only did a few like that. I figured Inno should pick the colours for her characters, I just put a few in as examples of how I'd do it. If she hates it though, I can always make a cut-back version that only affects the player.

WoefulWombat:
Ah, I see. One problem with that would be grey, black and other dark colours, they'd be a nightmare for my eyes. Another would be that clothing most often doesn't define a character, so it's a little strange to me to make their speech colour dependant on that. Could tie it to just hair or skin colour, like you did with the cultists, but the first problem would then still persist.

A-Master-of-Puppets:
Well, the way it works is actually the other way around, NPCs pick favourite colours when they're generated and then take clothing from that. If you've ever fought an NPC multiple times and taken just one item of clothing from them each time, you've probably seen what a mess their outfit ends up; this is also intended to fix that. I'll add another colour list for the primary colour that excludes black (grey is quite readable), because yes, that might be an issue.

WoefulWombat:
Yeah that sounds like a good improvement on its own, even without linking the speech colour to it.

src/com/lilithsthrone/game/character/GameCharacter.java
```diff
@@ -1069,6 +1099,60 @@ public static void loadGameCharacterVariablesFromXML(GameCharacter character, St
 					nameElement.getAttribute("nameFeminine")));
 		}
 		
+		// Colours:
+		if(element.getElementsByTagName("colours").getLength() > 0)
+		{
```
picobyte: maybe
```
private String[] colorTypes = {"primaryColour", "secondaryColour", "underwearColour", "leatherColour", "denimColour", "metalColour"};
for (String ctp: colorTypes)
{
	try
	{
		character.secondaryColour = Colour.valueOf(colourElement.getAttribute(ctp));
	}
	catch(IllegalArgumentException e)
	{
		System.err.println("Warning: Invalid " + ctp);
	}
}
```

CognitiveMist: I'm not sure how to make that work, since we are setting six different variables here.

CognitiveMist:
@A-Master-of-Puppets 

Innoxia:

PR #246, opened 2018-03-17
Remove requirement for characters to have breasts to pinch nipples

A-Master-of-Puppets:
Arguably, you should be able to grope someone's chest too even if they're completely flat, but I think that would require some rewriting of the descriptions. This was a purely mechanical change, which I'm more comfortable with.


Innoxia:
Thank you for catching this! I agree that the "grope breasts" action could be changed to include "grope chest", but yeah, I'd need to write in variations for that. ^^

I'll add it as an issue. :3

PR #249, opened 2018-03-18
Additional hair styles

WoefulWombat:
I hope it's only those three files I need to change? Maybe I missed something, but nothing seems to crash on me. I don't think the blurbs from game/character/body/Hair.java are used anywhere, there aren't any potions for changing hair styles and Kate doesn't use them either.

**Nine new hair styles:** Topknot, hime cut, chonmage, dreadlocks, twin braids, crown braid, drills, low ponytail and a bird cage.

The bird cage is a joke for people who played Dragon Age: Origins and romanced Leliana as a female PC. I wouldn't be mad if you axe that one, but with floor-length hair you already look ridiculous, a bit more can't hurt.

Not sure if dreadlocks should be an uncommon hair style, but since afro, sidecut and Mohawk are, I put them there anyway.


Innoxia:
Thank you for this! :3

PR #251, opened 2018-03-18
Maven resources fix and instructions

pinnae:
This is to compensate for adding the new `/res` folder a couple of versions back and fix the other path which I think I wrote wrongly
I've also added some instructions for using Maven to the build tutorial


Innoxia:
Thank you! ^^

PR #252, opened 2018-03-18
Fix speech parsing by femininity

A-Master-of-Puppets:
Creating a line of speech by femininity would only use the regular masculine or feminine colours, even if fed a plus version of masculine or feminine; now all five femininity variants work fine for that purpose.


Innoxia:
Thank you! :3

PR #259, opened 2018-03-19
Tail actions

A-Master-of-Puppets:
I've added the tail actions to all the positions that didn't have them, but individually this time. I have a hard time imagining a position that could preclude tail actions and still be described as "sex", but I suppose it could happen.


WoefulWombat:
I suppose it depends entirely on the length of the tail and/or your imagination and how much control you have over it. That's something that can certainly be expanded upon in the future—tail length, thickness, and modifiers. 
For instance, IRL cats don't have that much control over their tails, they are not "prehensile" by definition as they can't grasp with it. Very few animals can. So being able to fuck someone with a tail like that makes little sense. unless you say that catgirls do have that extra control and add a "prehensile"-modifier to it.
And then what about shorter tails? Anything below torso length is going to be awkward in 20-30% of all positions.

A-Master-of-Puppets:
Well, tail-fucking is already a thing in the game; it's not something new I'm adding. All I'm doing is enabling the actions for the positions that didn't have them available.

WoefulWombat:
Yeah, I know, I was just rambling about it in general. Should've just written that as a suggestion.

Innoxia:
I've completely changed how the game's sex positions get appropriate sex actions, so this PR is no longer valid. ^^

Thank you very much for your effort though; I did consult this when changing the sex position code. :3

Issue #204, opened 2018-03-11
Character creation bug


Innoxia:
When you attempt to start up a new game while playing an existing game, if you choose a female character is doesn't want to let you pick the androgynous option. If you want your character to be a tomboy from the start, the game won't let you without closing out of the game and restarting.

bicobus:
Issue related to femininity and gender: if you pick a male start then an androgynous femininity, you can't grow a beard. It seems that the femininity trumps biology.

Innoxia:
Fixed by Master of Puppets.

PR #263, opened 2018-03-20
Set BodyChanging target to player earlier in character creation

A-Master-of-Puppets:
BodyChanging target needs to be set to the player _before_ gender selection, fix for #204.


Innoxia:
Thank you! ^^

PR #267, opened 2018-03-20
Fix for loading the game advancing time

A-Master-of-Puppets:
Previously, loading the game would advance time from the point the game was saved.


Innoxia:
Thank you! :3

Issue #268, opened 2018-03-20
Squirter not saving?


Innoxia:
"I do not know if it is a bug or a problem with my setup but when every I add squirter to one of my characters, after a short time the trait removes itself."

Innoxia:
Fixed by Master of Puppets

PR #269, opened 2018-03-20
Fix vagina not saving squirter status

A-Master-of-Puppets:
Fix for #268.


Innoxia:
Thank you! ^^

Innoxia:
I might need to wrap the load in a try-catch...

A-Master-of-Puppets:
I tested that, Boolean.valueOf() returns false on a null.

PR #277, opened 2018-03-21
Clothing can affect apparent body part size

CognitiveMist:
I had an idea for this, so I made the PR as a proof of concept / phase one.

This change adds support for clothing to influence percieved breast and penis size, which affects the clothed character descriptions.

In this change, the looseness of the cloak reduces both parts (which helps obscure sex), the bagginess of the cargo trousers reduces bulge, and the tightness of a chest sarashi reduces cup size. Future clothing items, such as a push-up bra or padded underwear could also increase the apparent size.

Also includes adjusted penis reveal descriptions for when the penis size was significantly reduced. No adjustments were made to the breast reveal, since they were worded in a way that doesn't imply an expected size beyond e.g. masculinity.


A-Master-of-Puppets:
Can I just whine for a second about you including a bunch of typo fixes in your commit? Having to scroll through those to get to the mechanical changes is annoying.

With that out of the way: Does this actually work? It looks like it shouldn't, as at the point the cock is revealed, there isn't anything covering it anymore and getAllLayersCoverableArea should return an empty set.

CognitiveMist:
Hmm, you may be right about the reveal text. I simplified that at the end and and may have broken it.

The sex/gender-prediction within the character description definitely does work, however.

Re: the grammars changes, I was going to reword those like I did for the penis, but decided my changes didn't add anything, I'm no writer, but I left the grammar fixes in. Sorry for your inconvenience.

A-Master-of-Puppets:
It's not a big deal, I just think you should try to keep one submission doing one thing. The grammar fixes would have been better in a separate PR.

It might still be a little weird, maybe? I mean, the true size of your bulge is going to be visible when you take your trousers off, so why should it be a huge surprise later on when you remove your underwear? I like the idea of the sarashi reducing breast size, certainly, but I think there are some issues with doing the same with cocks and baggy trousers.

CognitiveMist:
I see your point, but I'm not sure how to do better without a much more invasive change. Should they instead voice their surprise when the bulge becomes visible (i.e. check on every clothing removal)? That seems klunky from a gameplay point of view to me.

Perhaps this could be solved with better writing in the responses, something like "that bulge wasn't there five minutes ago!", or something.

We don't really have any compressive underwear yet (like the sarashi, but for the crotch). I suppose the chastity cage/belt could do so, but usually those aren't removed during sex.

A-Master-of-Puppets:
I'm not sure what's wrong with them reacting when the bulge becomes visible; that seems like when they logically _should_ react to me.

CognitiveMist:
Hmm, I suppose, but if multiple layers have affect the body part, that might cause them to trigger a reaction at each step. "Oh, you have a bulge" "Oh, you have a bigger bulge" "Oh, you have a bigger bulge" "There it is". This could work, but it might break the flow of the other stuff that's going on.

I could give that a try, but I feel like it might need a writer's pass to make it more clear what is happening. We'll see.

mirh:
Once you/NPC figure out there's a sizeable/noticeable/whatever bulge *once*, I don't see what are the other concerns?

Also, put nonetheless aside the very smart "clothing affects apparent body part size" thing.. The real disheartening thing is that you can even have a python in a thong, and nobody will bat an eye until the very second you pull it off.

CognitiveMist:
Well, I kind of forgot about this. I think this is a good first step, conceptually, but there's a lot more that should be there; such as surprise the first time, but not after, or surprise if you've tf'd since last time. That would be a pretty heavy-handed change, though.

CognitiveMist:
I'm going to abandon this since I haven't had any motivation to work on it further.

PR #278, opened 2018-03-21
Fixed addiction time bug

### LINKED ISSUE #276: NPC addiction logic faulty
When you give an addicted NPC the substance they are craving it always says that they weren't suffering from withdrawal, even if they have the withdrawal status effect.

Edit: Further testing shows this applies to the PC as well.

AlacoGit:
This fixes #276


Innoxia:
Thank you! ^^

PR #279, opened 2018-03-21
Adjust the display of subspecies lists in the Encyclopedia.

tukaima:
Currently, subspecies names/info are displayed in two separate lists in the Encyclopedia. The first is contained within the same "frame" as all the other text, and the other in the average stats display.

The former has been hidden underneath an HTML details element, so that the player is not greeted with ten thousand rows of text when they open the Slime page (and any future race entries which have a similar amount of subspecies).

The latter is superfluous, so it has been removed; the second Subspecies display shares no information that is not already available in the first. You could make a case for it being easier to read through, but that's about it.


Innoxia:
Thank you for this! As there's a merge conflict, I quickly implemented it in my local code (and credited you).

PR #280, opened 2018-03-21
Renames Dobermann-morphs

Rfpnj:
Changes the morph name to Dobermorphs, well keeping the pronoun as dobermann. This would only effect the times when you see them in battles, or are going for the subspecies your self. This was suggested by Jimmythehand 3/21/18 at about 12:31pm EST


Innoxia:
The term "dobermann-morph" is rather clunky, I agree. Instead of Dobermorphs, however, I wonder if it would be better to simply call them Dobermann/Dobermanns?

Innoxia:
Ok, I'll change them to "Dobermann/Dobermanns". :3

PR #285, opened 2018-03-23
Add Companion framework base.

Irbynx:
Adds the groundwork for the companions/followers to the game. Doesn't include followers themselves, however, most of the functionality is here to create a sufficiently decent follower already, barring a few placeholder texts.

- Companions are GameCharacters. Any character can be a companion, although for obvious reasons the behavior might be a little weird, especially for slaves.
- Companions follow their PartyLeader around and join them in most instances of combat. The only combat the companions won't automatically join is combat initialized with one of the combat dialog overloads that accepts two sides of combat as arrays, which I assume is going to be used to have some special combats going on.
- Rudimentary support for scheduling added. NPCs can override their handleSchedule() function which will be called during the main turn end loop to do their scheduled actions. For example, it will allow some NPCs that the player can meet outside to only show up during the day while there is no arcane storm going on.
- Companions button added to phone menu; only available if player has an NPC following them. Opening it they can choose one of their followers to interact with. For now interactions are limited to "Go Home" and "Sex" but stuff like "Spar" and "Inventory" should probably be added too.
- Sex menu includes dominant and submissive scenes along with a support for threesomes with your other companions (with Spitroast and Side-to-Side pose). Companions need to at least be indifferent to dominant or submissive sex respectively to have the dom/sub options to show up, and in addition to that the player needs to overload their getCompanionSexRejectReason() function to allow it give proper reasons as to why they won't do sex (or return empty strings if they would), i.e no exhibitionistic sex in public for shy characters and so on. As expansion, could probably include rape sex actions if the player is stronger than them.
- Companions will automatically ditch the player if their requirements aren't met during a turn update (set in the respective function).
- Morality checks. Very rudimentary, but allows the NPCs to react to things they don't like and ditch the player once they do something bad or get more affection if they do something good. In their eyes, obviously (right now I've added stubs for Violence, Slavery, Surrendering and Dishonesty, but more can be added obviously).
-  Companion state is saved and loaded properly... I hope.
- Arkhi, a future lynx-morph companion (cat-morph for now) is in files as well; used for debugging. To enable her, uncomment the code in initNewGame and in Lilaya's garden.
- Any GameCharacter can be a companion, although player support is limited; a player as a companion will not follow an NPC, has no real restrictions (aside from other NPCs being unable to join them as companions) and doesn't save the companion state to XML.

All in all, nothing that the end user should be excited about yet, lol. 

Outstanding issues:
- Game treats all the characters that aren't a player as slaves in the areas where the slaves are supposed to be. Which means that your companion will immediately drop down to the floor to start polishing it and you can interact with them like a slave...
- I also didn't manage to get an alleyway attacker to jump on me while exploring the alleyways. It showed my companion as the one who prowls the area, however. The lust storm attackers resolve as normal.
- During the combat, the allied NPCs still have their attacks say that they hit the player. Technically, they don't, but this will definitely confuse people.


Innoxia:
Thank you so much for this! I'll look through it and try to merge it over the weekend! :3

Innoxia:
This is a lot for me to go through and check, but I will merge this soon! ^^

Irbynx:
I'm waiting with hype engine revved at max :D

Innoxia:
Thank you for this! It will be merged into the github update tomorrow (I've merged it in my local code already). :3
I've removed the part about Arkhi for now, as it would be best to add her as a full NPC all in one go.

Irbynx:
Allright, that was my plan to begin with (although I'd need the catmorph PR to be in for that too) :D

PR #290, opened 2018-03-25
Punctuation fixes

WoefulWombat:
Dialogue tags!
After looking it up, it seems the rules my old English teacher taught us still hold true. So, no commas for narration, dialogue tags that sit between two independent clauses end on a fullstop etc. [This thing](http://theeditorsblog.net/2010/12/08/punctuation-in-dialogue/) summarises it nicely.
The question is though what you constitute as a dialogue tag, people seem to argue about that. 
This is just my own opinion, but things like "he smiles" or "she frowns" are not tags, while "he laughs/growls/sighs" can be, depending on context. Like, you can laugh out words or growl them, but you can't smile a sentence. So anything that affects the words, the tone, intonation or pace, can be a tag. But it could also not be, if it is a separate action. That's just my two cents, there are enough people who are far stricter and just see "say", "ask" and their variations as dialogue tags.

I only went through the Nyan content, so this is what it would look like. Might have missed some stuff though, I'm a little rusty. Some cases are also ambiguous and would be correct either way. What are your thoughts on the matter?

Again, rules are all fine and dandy, but being consistent is most important. Nothing takes the reader easier out of the flow than jarring differences.

Ah, and in case you do decide to do it this way, lemme know. I can go through the rest of the game and quickly do the thing, so it doesn't look too jarring by being the only part with different punctuation.
Otherwise, just ignore it. Consistency above all else.

### DIFF
--- a/res/txt/characters/dominion/nyan.xml
+++ b/res/txt/characters/dominion/nyan.xml
@@ -28,8 +28,8 @@
 	<htmlContent tag="NYAN_TALK"><![CDATA[
 	<p>
 		[pc.speech(I was wondering how your work's been going recently,)]
-		 you reply,
-		[pc.speech(your store looks as busy as ever today.)]
+		 you reply.
+		[pc.speech(Your store looks as busy as ever today.)]
 	</p>
 	<p>
 		[nyan.speech(W-Well, it's b-been going well, t-thank you,)]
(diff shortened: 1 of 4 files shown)


Innoxia:
Thank you very much for taking the time to do all of this, and for letting me know how to properly format dialogue! I've made use of this information in the new scenes for Amber (version 0.2.3), adn will continue to format dialogue as you've shown me from now on. ^^

PR #294, opened 2018-03-25
Fix inserted bimbo/sex/muffled/drunk words breaking capitalisation

pinnae:
I simplified down some of the logic and added a single method `insertIntoSentence` that does most of the stuff that's repeated in those other methods.
When words are inserted at the beginning of sentences it now lowercases the following word that was originally the start of the sentence.
An edge case I'm not sure how to deal with is if sentences start with names or titles (like your nickname).

Edit: Maybe a simpler method to avoid unwanted decaps would be to avoid inserting into the start of a sentence entirely? If correcting caps is wanted, this seems like the only reasonable way to ensure correctness.

### DIFF
--- a/src/com/lilithsthrone/utils/Util.java
+++ b/src/com/lilithsthrone/utils/Util.java
@@ -538,9 +538,9 @@ public class Util {
 	 * Determine whether a given string contains sentence-ending punctuation
 	 *
 	 * @param text
-	 * 						text to check whether
+	 *            text to check whether
 	 * @return
-	 * 						boolean, whether the text contains a period, exlamation or question mark
+	 *            boolean, whether the text contains a period, exlamation or question mark
 	 */
 	private static boolean isEndOfSentence(String text) {
 		return endOfSentence.matcher(text.substring(text.length()-1)).matches();
@@ -550,16 +550,16 @@ public class Util {
 	 * Inserts words randomly into a sentence.</br>
 	 *
 	 * @param sentence
-	 *						sentence to insert words into
+	 *            sentence to insert words into
 	 * @param frequency
-	 *						how often words are inserted. 1/frequency is the probability of inserting a word
+	 *            how often words are inserted. 1/frequency is the probability of inserting a word
 	 * @param inserts
-	 *						list of strings to insert into. These are appended to the end of words, so ensure
-	 *						any whitespace wanted is put before the insert. A space separates the next word
+	 *            list of strings to insert into. These are appended to the end of words, so ensure
+	 *            any whitespace wanted is put before the insert. A space separates the next word
 	 * @param middle
-	 *						boolean, whether to avoid inserting at the start/end of a sentence
+	 *            boolean, whether to avoid inserting at the start/end of a sentence
 	 * @return
-	 *						modified sentence
+	 *            modified sentence
 	 */
 	private static String insertIntoSentences(String sentence, int frequency, String[] inserts, boolean middle) {
 		splitSentence = sentence.split(" ");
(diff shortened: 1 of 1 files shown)


src/com/lilithsthrone/utils/Util.java
```diff
@@ -524,41 +565,9 @@ public static String addStutter(String sentence, int frequency) {
...
-
-		int offset = 0;
-		for (int i = 0; i < wordsToBimbofy; i++) {
-			offset = random.nextInt(frequency);
-			offset = ((i * frequency + offset) >= splitSentence.length ? splitSentence.length - 1 : (i * frequency + offset));
-			if (offset != 0) {
-				// If previous word didn't end with punctuation:
-				if (splitSentence[offset - 1].charAt(splitSentence[offset - 1].length() - 1) != '.' && splitSentence[offset - 1].charAt(splitSentence[offset - 1].length() - 1) != '!'
-						&& splitSentence[offset - 1].charAt(splitSentence[offset - 1].length() - 1) != '?') {
-					// Add a comma to the end of the previous word:
```
CognitiveMist: Unless I'm missing something, the previous-word comma added here isn't in the new function. Overall, I like the change.

CognitiveMist: Unless I'm missing something, the previous-word comma added here isn't in the new function.

src/com/lilithsthrone/utils/Util.java
```diff
@@ -504,6 +504,47 @@ public static String addStutter(String sentence, int frequency) {
...
 
+	/**
+	 * Inserts words randomly into a sentence.</br>
+	 * 
+	 * @param sentence
+	 *            sentence to insert words into
+	 * @param frequency
+	 *            how often words are inserted. 1/frequency is the probability of inserting a word
+   * @param inserts
+   *            list of strings to insert into
```
CognitiveMist: Nitpick, but whitespace consistency.

src/com/lilithsthrone/utils/Util.java
```diff
@@ -504,6 +504,47 @@ public static String addStutter(String sentence, int frequency) {
...
+
+		// 1 in "frequency" words are muffled interjections, with a minimum of 1.
+		int wordsToInsert = splitSentence.length / frequency + 1;
+
+		int offset = 0;
+		for (int i = 0; i < wordsToInsert; i++) {
+			offset = random.nextInt(frequency);
+			offset = ((i * frequency + offset) >= splitSentence.length ? splitSentence.length - 1 : (i * frequency + offset));
+
+			// Decapitalise the current word if it was originally the start of a sentence and isn't all caps
```
CognitiveMist: Where is the beginning of the sentence re-capitalized?

pinnae:
I've made changes since, changing the way I've done some things
`insertIntoSentences` has an optional `middle` arg that lets you choose if you want to allow the inserts to happen at the start or end of a sentence at all
This avoids the problem of working out if a word can be decapitalized, but it's still an imperfect solution
Inserts are now added to the _end_ of words, so spacing has been fixed for `addBimbo` and the rest

Before (Bimbo):
> Hi Master! Like, Oh, I'm so happy to, um, see you again! I've been on my best, like, behaviour!

After (Bimbo):
> Hi Master! Oh, I'm so happy, ah, to see you again! I've, uh, been on my, like, best behaviour!
After (Bimbo+Drunk):

> Heeey Master! Oh, I'm sho \~Hic!\~ happy, uh, to \~Hic!\~ see you again! I've been, ah, on \~Hic!\~ my, like, best behaviour and stuff! \~Hic!\~.

From my testing with a higher insert frequency there's no case where this breaks

Innoxia:
Thank you very much for this! <3

Issue #283, opened 2018-03-22
Save-able templates for enchanting


A-Master-of-Puppets:
This has been requested a whole bunch, but it doesn't have an issue for it (it's mentioned in #134, but it should have one for itself). I've been thinking of having a go at this, but I'm not sure where I should save the templates _to._ Do they want to go in the save file, or in the properties file?

AlacoGit:
Of the existing files the properties file makes the most sense since recreating the templates between saves might get annoying.

(On a side note: Might be worth looking into import/export for these templates if the enchanting system gets expanded on)

bicobus:
> Do they want to go in the save file, or in the properties file?

As a single player could run several play through, I would argue for a place distinct from the save file. I sometimes nuke my properties.xml file, beside one file per potion could allow for an out of game creation of templates.

TadUnlikely:
@Innoxia @A-Master-of-Puppets Can this be closed now?

AceXP:
I am going through some old issues. Since this one is resolved, could you close it?

Only Inno and the owner of the issue can close issues, and Inno's time is better spent on development.

Thanks!

PR #296, opened 2018-03-27
Saveable recipes for enchanting

A-Master-of-Puppets:
As discussed in #283, this adds a screen to the enchanting UI where you can save/load the current list of enchanting effects to files. There is one issue, since the effect type must match between the item type being enchanted and the item type in the file, you can't load a potion saved using the "Sexual" type potion when enchanting a "Corruptive" type potion, as these are technically different ItemEffectTypes even though all effects a "Sexual" type potion can have, a "Corruptive" type potion can _also_ have. Likewise, a transformation potion saved using a cow-morph item can't be loaded to enchant a dog-morph item even if you aren't using any cow-specific effects. I'll be looking at ways to fix this soon, I hope.

### DIFF
--- a/src/com/lilithsthrone/controller/MainController.java
+++ b/src/com/lilithsthrone/controller/MainController.java
@@ -762,6 +762,17 @@ public class MainController implements Initializable {
 								}
 							}
 						}
+						if(Main.game.getCurrentDialogueNode() == EnchantmentDialogue.ENCHANTMENT_SAVE_LOAD){
+							if((boolean) Main.mainController.getWebEngine().executeScript("document.getElementById('new_save_name') === document.activeElement")) {
+								allowInput = false;
+								if (event.getCode() == KeyCode.ENTER) {
+									enterConsumed = true;
+									Main.mainController.getWebEngine().executeScript("document.getElementById('hiddenPField').innerHTML=document.getElementById('new_save_name').value;");
+									EnchantmentDialogue.saveEnchant(Main.mainController.getWebEngine().getDocument().getElementById("hiddenPField").getTextContent(), false);
+									Main.game.setContent(new Response("Save", "", Main.game.getCurrentDialogueNode()));
+								}
+							}
+						}
 						if(Main.game.getCurrentDialogueNode() == SlaveryManagementDialogue.SLAVE_MANAGEMENT_INSPECT
 								|| Main.game.getCurrentDialogueNode() == SlaveryManagementDialogue.SLAVE_MANAGEMENT_JOBS
 								|| Main.game.getCurrentDialogueNode() == SlaveryManagementDialogue.SLAVE_MANAGEMENT_PERMISSIONS){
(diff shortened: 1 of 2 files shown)


src/com/lilithsthrone/game/dialogue/utils/EnchantmentDialogue.java
```diff
@@ -418,4 +444,296 @@ public static void initModifiers(AbstractCoreItem ingredient) {
...
+			transformer.setOutputProperty(OutputKeys.INDENT, "yes");
+			transformer.setOutputProperty("{http://xml.apache.org/xslt}indent-amount", "2");
+			DOMSource source = new DOMSource(doc);
+			
+			String saveLocation = "data/enchantments/"+name+".xml";
+			StreamResult result = new StreamResult(new File(saveLocation));
+			
+			transformer.transform(source, result);
+			
+		} catch (ParserConfigurationException pce) {
```
Pimvgd: you can use multi-catch here

src/com/lilithsthrone/game/dialogue/utils/EnchantmentDialogue.java
```diff
@@ -418,4 +444,296 @@ public static void initModifiers(AbstractCoreItem ingredient) {
...
+					effects = effectsToBeAdded;
+				}
+				catch(Exception e)
+				{
+					e.printStackTrace();
+				}
+			}
+		}
+	}
+	public static boolean isLoadEnchantAvailable(String name)
```
Pimvgd: make a private static Optional<File> which returns empty optional if the file doesn't exist, and returns the file handle if it does. You can then use it in deleteEnchant, and in loadEnchant with ifPresent.

src/com/lilithsthrone/controller/MainController.java
```diff
@@ -5148,6 +5159,88 @@ public void effects() {
 			}
 		}
 		
+		// Save/load enchantment:
+		if (Main.game.getCurrentDialogueNode() == EnchantmentDialogue.ENCHANTMENT_SAVE_LOAD) {
+			for (File f : EnchantmentDialogue.getSavedEnchants()) {
+				id = "overwrite_saved_" + f.getName().substring(0, f.getName().lastIndexOf('.'));
```
Pimvgd: you make use of "f.getName().substring(0, f.getName().lastIndexOf('.'))" a lot here. Recommend assigning to local variable.

A-Master-of-Puppets:
OK, the new check to see if imported ItemEffects are allowed will correctly handle different ingredient types so long as the primary and secondary modifiers match. This fixes the issue mentioned in the first post.

I'll have a look at the suggested refactors now; initially I just copy/pasted the code from the save/load game methods for expediency.

Innoxia:
Thank you so much for this! I've merged and expanded upon this in my local branch, so this will be merged for 0.2.5's release. ^^

PR #298, opened 2018-03-27
Fix for receiving oral for stocks

### LINKED ISSUE #260: Unable to receive oral sex in the public stocks
This is kind of strange, but I believe that the *slave* must be allowing vaginal use in order for a PC with only a vagina to receive oral sex.

Reproduce:
PC with only a vagina
NPC not allowing vaginal use (or with no vagina)
Choose Position -> Move to Front
No Sex actions available.

A-Master-of-Puppets:
This should fix #260. The method is question is buried kind of deep in the guts of the engine, so I can't be 100% sure there are no side-effects to this fix. I don't _think_ there should be, though.


Innoxia:
Thank you for the PR! I've fixed this on my end (it took a little more to get it all working properly), but your PR was a really useful starting point! I'll credit you for this. :3

PR #299, opened 2018-03-28
Make cum addicts enjoy swallowing cum

A-Master-of-Puppets:
This is a minor text change for when the PC takes a cumshot down their throat and should want it.

### DIFF
--- a/src/com/lilithsthrone/game/sex/sexActions/baseActionsMisc/GenericOrgasms.java
+++ b/src/com/lilithsthrone/game/sex/sexActions/baseActionsMisc/GenericOrgasms.java
@@ -1220,40 +1220,103 @@ public class GenericOrgasms {
 				case MOUTH:
 					if(target.isPlayer()) {
 						cumTargetSB.append(" deep down your throat, and you find yourself making muffled whining noises as you feel the [npc.cum+] sliding down into your stomach.");
-						switch(characterOrgasming.getCumFlavour()) {
-							case BEER:
-								cumTargetSB.append(" The unusual taste of [npc1.name]'s beer-flavoured cum rises up to hit your [npc2.tongue], and you're left with no other option but to gulp down as much of the");
-								break;
-							case CHOCOLATE:
-								cumTargetSB.append(" The unusual taste of [npc1.name]'s chocolate-flavoured cum rises up to hit your [npc2.tongue], and you're left with no other option but to gulp down as much of the");
-								break;
-							case CUM:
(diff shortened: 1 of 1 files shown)


src/com/lilithsthrone/game/sex/sexActions/baseActionsMisc/GenericOrgasms.java
```diff
@@ -1220,40 +1220,103 @@ private static String cumTargetDescription(GameCharacter characterOrgasming, Gam
...
-								cumTargetSB.append(" The unusual taste of [npc1.name]'s strawberry-flavoured cum rises up to hit your [npc2.tongue], and you're left with no other option but to gulp down as much of the");
-								break;
-							case VANILLA:
-								cumTargetSB.append(" The unusual taste of [npc1.name]'s vanilla-flavoured cum rises up to hit your [npc2.tongue], and you're left with no other option but to gulp down as much of the");
-								break;
+						if(target.hasFetish(Fetish.FETISH_CUM_ADDICT) || SexFlags.playerRequestedCreampie)
+						{
+							switch(characterOrgasming.getCumFlavour()) {
+								case BEER:
+									cumTargetSB.append(" The unusual taste of [npc1.name]'s beer-flavoured cum rises up to hit your [npc2.tongue], and you "
```
pinnae: These switch statements seem to end in the same format so how about something like this

```java
switch(characterOrgasming.getCumFlavour()) {
    case BEER:
        cumTargetSB.append(" The unusual taste of [npc1.name]'s beer-flavoured cum rises up to hit your [npc2.tongue], and you");
    // all other cases
}

if(target.hasFetish(Fetish.FETISH_CUM_ADDICT) || SexFlags.playerRequestedCreampie) {
    cumTargetSB.append(" " + UtilText.returnStringAtRandom("greedily","hungrily") + " gulp down as much of the");
else {
    cumTargetSB.append("'re left with no other option but to gulp down as much of the");
}

A-Master-of-Puppets: That would fragment the lines more, though. I worry that that would make it hard to follow.

CognitiveMist: I agree with pinnae here; the lines are already quite fragmented.

If you want to de-frag the flow, I'd suggest splitting it out into a separate method for each chunk, then make cumTargetDescription() just stitch things together. You might have to do this for other descriptions, to make things consistent, though.

Innoxia:
Great stuff; thanks to both of you for this! <3

Issue #221, opened 2018-03-15
Sex Action arousal increases need looking at


Innoxia:
Some actions shouldn't affect arousal? #193

A-Master-of-Puppets:
There are literally only five actions that decrease arousal in the game:
- Pix denying the player
- The player calming down
- The player denying an NPC
- The two general "Resist" actions
All the actions to resist a specific action, like "Resist Oral", and also the "Beg for [himHer] to stop" dirty talk variant, these all have _positive_ arousal.
I would suggest that the "Resisting" actions should probably cause zero arousal, with the exception of ones that involve stimulation such as being fucked or licked/blown against their will.
Also, restrict the player from using "Calm Down" if they are being raped; being able to slow down and take a breather should require some measure of control of the scene.
That would fix the problem where it is very difficult to make an NPC cum against their will, and they simply can't do it _at all_ to the player.

Innoxia:
I've fixed this for 0.2.3.5. :3

PR #300, opened 2018-03-28
Zero arousal changes from resisting actions

A-Master-of-Puppets:
As discussed in #221, this sets all actions in the resisting pace that don't involve direct stimulation to have zero arousal gain. This prevents the standard "resist" action from making it virtually impossible to make an unwilling character cum. This also prevents the player from using the "Calm Down" action while being raped, for the same reason.

### DIFF
--- a/src/com/lilithsthrone/game/sex/sexActions/SexActionUtility.java
+++ b/src/com/lilithsthrone/game/sex/sexActions/SexActionUtility.java
@@ -88,6 +88,16 @@ public class SexActionUtility {
 		}
 
 		@Override
+		public boolean isBaseRequirementsMet()
+		{
+			if(Sex.isDom(Main.game.getPlayer()))
+				return true;
+			if(Sex.isConsensual())
+				return true;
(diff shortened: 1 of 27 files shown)


Innoxia:
Thank you for the PR! Wouldn't it be a lot easier to just put the resisting check in SexAction's getArousalGain methods? I'm going to implement it like that on my end now, but I will use your work to fix the other negative arousal gain instances in the game.

Innoxia:
I've implemented this in my local code, but will credit you in the patch notes. Thank you for your work! ^^

A-Master-of-Puppets:
The reason I did it on an action-by-action basis was so that some actions (the ones involving actual fucking) would still cause arousal gain. Thinking about it, that _could_ be done in the super method since actions have orifices associated with them. I kind of missed that at the time. Oh well.

Innoxia:
Yeah, I put in a check for orifices in the super method. :3
Thank you again for your work on this! ^^

PR #302, opened 2018-03-29
Add one-shot actions to lubricate tails

A-Master-of-Puppets:
Similar to how you can with fingers.


Innoxia:
Looks great, thank you! ^^

PR #303, opened 2018-03-29
Add management tab to slave interface

A-Master-of-Puppets:
This adds a management tab to the dialog when talking to a slave. It's a little quirky, as the "back" buttons on the management dialogs dump you back to the main room dialog rather than the slave dialog, but other than that it seems to work fine.


Innoxia:
A very welcome improvement to the slave system; thank you very much! ^^

Issue #270, opened 2018-03-20
Rework sex to include position lists for scenes?


A-Master-of-Puppets:
So I went to try and add some positions to some of the non-standard scenes (witches, chair sex, etc.), and it looks like scenes don't have a list of available positions? I'm having a hard time wrapping my head around that; it looks to me like the only way to add a new position to, for example, the witches' scene is to completely duplicate it as something like SMCowgirlCultist, and add position change actions manually? Am I misunderstanding this? What I would expect is to have a scene object, like SceneCultist or SceneMugger, which contains a set of available positions, and to add a position you would just add the position object to that.

TadUnlikely:
@Innoxia I believe this is fixed now across the board.

PR #305, opened 2018-03-30
Chair scene upgrades

A-Master-of-Puppets:
This just adds a few actions to the chair scenes that were missing. Main change is adding fingering so lesbian characters can actually _do_ something in these scenes, but also allow PCs with tails to use these.


Innoxia:
I really need to update this entire scene fully to the "new" system. This was the very first sex scene I added to the game (as a test), and it still uses some of the obsolete framework from all the way back in version 0.1.0.

Your changes are very welcome, so thank you very much for these! ^^

A-Master-of-Puppets:
I considered doing more (I particularly want to add a position for kneeling/seated oral), but adding new positions seems really clunky at the moment. I think the system should probably be upgraded with something like I suggest in #270, which would make it a lot easier.

PR #306, opened 2018-03-31
Block clothing displacement for Ralph's scene

A-Master-of-Puppets:
This prevents Ralph from stripping you, or you from stripping him, during his scene.


Innoxia:
Good idea -  the clothing already gets automatically displaced when moving to the 'Big discount' scene, so I can't see this causing any issues.

Issue #310, opened 2018-04-01
"Stop Sex" Does not return to any menu


tyr4n02r:
Stop sex (having tried Multiple, Slave, Rape and Pix) does not return me to the previous menu (World Map, post-fight, etc.) Issue as of 2.2. Also on .jar if that helps anything.

A-Master-of-Puppets:
Can you attach your error log?

tyr4n02r:
any instructions no how to do that? Sorry not a programmer, was just reporting the error.

A-Master-of-Puppets:
Look for the error log file in the data folder, and attach it like it says below the box where you type in your message here.

tyr4n02r:
/Users/Admin/Desktop/Sylenth1 Soundbanks/Fenoxo/Lilith's Throne/data/error.log

tyr4n02r:
Take 2...

tyr4n02r:
Alright, seriously...
Its not uploading to the site unless you guys see something on here that I dont.

tyr4n02r:
Yea. The file isnt upload ing either... 8 attempts with different browsers...

tyr4n02r:
Ok so since it doesnt seem that big ill post the code. Sorry for the dump.

java.lang.IllegalArgumentException: The supplied clothing cannot be displaced in this manner!
	at com.lilithsthrone.game.inventory.CharacterInventory.isAbleToBeDisplaced(CharacterInventory.java:1288)
	at com.lilithsthrone.game.inventory.CharacterInventory.isAbleToBeDisplaced(CharacterInventory.java:1217)
	at com.lilithsthrone.game.character.GameCharacter.isAbleToBeDisplaced(GameCharacter.java:9924)
	at com.lilithsthrone.game.sex.Sex.endSex(Sex.java:479)
	at com.lilithsthrone.game.sex.Sex.access$7(Sex.java:462)
	at com.lilithsthrone.game.sex.Sex$1$1.effects(Sex.java:1049)
	at com.lilithsthrone.game.dialogue.responses.Response.applyEffects(Response.java:174)
	at com.lilithsthrone.game.Game.setContent(Game.java:1185)
	at com.lilithsthrone.controller.eventListeners.SetContentEventListener.handleEvent(SetContentEventListener.java:36)
	at com.sun.webkit.dom.EventListenerImpl.fwkHandleEvent(EventListenerImpl.java:129)
[...]

PR #313, opened 2018-04-05
Fix for bad displacement imports crashing after sex

A-Master-of-Puppets:
Fix for #310, checks that imported clothing displacements are valid on import.


Innoxia:
Thank you! ^^

Issue #322, opened 2018-04-09
Character images broken on non-windows OSes


A-Master-of-Puppets:
On OSes other than windows, file paths are case-sensitive, so `res/images/characters/nyan` (where the images are) is not the same place as `res/images/characters/Nyan` (where the game actually looks).

Innoxia:
This was fixed in #323

PR #323, opened 2018-04-10
Fix for broken images on non-windows

A-Master-of-Puppets:
This capitalises the names of the artwork folders, as mentioned in #322.


Innoxia:
Can this also be fixed by fixing the `Artwork` class's constructor? It was meant to convert the names to lowercase, but I forgot to put `this.name` in the File constructors.

Basically, just by changing `Artwork`'s name argument to be called nameInput, then changing line 86 to:
`this.name = nameInput.toLowerCase();`

https://github.com/Innoxia/liliths-throne-public/blob/master/src/com/lilithsthrone/rendering/Artwork.java

A-Master-of-Puppets:
That might work, but you'd still have to rename the folders for the harpies and Lilaya, since they have uppercase letters in them already. I think having the folders simply share the name of the classes is probably a better solution.

A-Master-of-Puppets:
OK, that definitely _doesn't_ work. The images still don't show up for most characters, and it actually crashes the game when I try to inspect Lilaya.

Innoxia:
Oh yeah, sorry, I didn't think of that! ^^

Thank you for this PR! <3

PR #325, opened 2018-04-13
Adjusts the floors in the post license content

Rfpnj:
The rooms were mentioning the first floor for self selling


Innoxia:
This is because I use the UK system of starting with "ground floor", and then the floor above that is the first floor. ^^

PR #326, opened 2018-04-13
Fix for Breast.setSize

A-Master-of-Puppets:
This allows Breast.setSize to accept a null value for the owner parameter, in a similar manner to how Ear.setType and Tail.setType work.


Innoxia:
Thank you! :3

PR #328, opened 2018-04-13
Fix new corset dress clothing item

A-Master-of-Puppets:
This fixes the new corset dress item. The main change is making it block the leg slot like other dresses do. Also includes reformatting the displacement section for consistency with other items (no functional changes) and replacing the ad-hoc colour lists with preset ones (although I question if it should really be restricted to black only by default).


Innoxia:
Thank you! I might add some more default colours. ^^

PR #330, opened 2018-04-14
Addiction patch 2

AlacoGit:
This changes ingestFluids to only show the "new addiction" dialog if the ingesting character isn't already addicted to the fluid, and to otherwise always show the "satisfied craving" dialog if the ingesting character is addicted.

This makes it a little cleaner to use the addiction system, since before if you fed an addicted character with an addicting version of a fluid, it would show the "new addiction" dialog every time.


Innoxia:
Thank you! :3

PR #332, opened 2018-04-14
Changes Nyan's quest from Side to Relationship, with all references fixed

Rfpnj:
Makes the required modifications to have quest tree listed as a romance quest. Tested and checked for compatibility. Change is not SAVE compatible. Tested on 0.2.5.5

### DIFF
--- a/src/com/lilithsthrone/game/character/quests/Quest.java
+++ b/src/com/lilithsthrone/game/character/quests/Quest.java
@@ -416,7 +416,7 @@ public enum Quest {
 	},
 	
 	
-	SIDE_NYAN_STOCK_ISSUES(QuestType.SIDE, 1, 0) {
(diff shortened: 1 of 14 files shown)


src/com/lilithsthrone/game/character/quests/Quest.java
```diff
@@ -416,7 +416,7 @@ public String getCompletedDescription() {
 	},
 	
 	
-	SIDE_NYAN_STOCK_ISSUES(QuestType.SIDE, 1, 0) {
+	RELATIONSHIP_NYAN_STOCK_ISSUES(QuestType.SIDE, 1, 0) {
```
Pimvgd: this... doesn't make sense. Why RELATIONSHIP_NYAN but QuestType.SIDE?

PR #342, opened 2018-04-16
Punctuation fixes, the finale

WoefulWombat:
And the actual rest of the dialogue punctuation, the .xml files, again including other small stuff I noticed. Adverbs ending on -ly are never followed by a hyphen. So something like "magically-enchanted" is wrong, it's "magically enchanted". Anything that's not an adverb and ending on a -ly is fine though, like "family-friendly" or "dilly-dally".


WoefulWombat:
Fixed 0.2.3.5 conflicts.

Innoxia:
Thank you!!!!! <3 <3 <3

PR #343, opened 2018-04-16
Rework repeat actions tab

A-Master-of-Puppets:
This reworks the repeat actions sex tab. Changes:
- Actions are now sorted by recentness, with most recent actions listed first
- Actions are no longer greyed out; they just don't appear
- Actions that are unavailable are not forgotten - for example, actions that require you to be fingering someone, like "Rough Fingering" or "Clit Focus" will no longer disappear from the recent actions tab if you stop fingering someone and then start again.


Innoxia:
Thank you! :3

PR #344, opened 2018-04-16
Rework harpy body selection

A-Master-of-Puppets:
Curious how harpy generation works, I discovered it's determined solely by femininity - this produced the situation where a player could set their gender preferences to 100% shemales or traps, and harpies would all generate as female. This doesn't seem right, if a player has a preference for cocks, they should _get_ cocks. This will fix that issue.


Innoxia:
Good catch - thank you very much for this!

PR #349, opened 2018-04-16
Fix [pc.cum+] to [npc.cum+]

ltpatch:


Rfpnj:
Please leave a comment on the PR in the future so that we can have an idea of what is changed without needing to check the files for it.

Good catch however

Innoxia:
Thank you! :3

PR #350, opened 2018-04-16
Creates PR template

Rfpnj:
Asks for three pieces of information with the intent of speeding up acceptance of future PRs, It will also alert Innoxia that a PR was submitted.


tukaima:
It may be worth including something such as "Are any new graphical assets required?".

Rfpnj:
Added additional question, and spaced out the lines for ease of reading.

Innoxia:
Thank you for this! :3

PR #352, opened 2018-04-17
Fix background colour for Harpy Nest entrance

A-Master-of-Puppets:
This just changes the background colour for the Harpy Nest entrance to match the darker colour the streets under the nests have.


Innoxia:
Thank you! ^^

PR #353, opened 2018-04-17
Typo fix

A-Master-of-Puppets:
You don't use an apostrophe for plurals, and you _certainly_ don't say "woman's" or "man's", which is happening currently.


Innoxia:
Good catch! ^^

PR #356, opened 2018-04-17
Gender utility methods

A-Master-of-Puppets:
This adds methods for getting the masculine, feminine, and androgynous equivalents for a given gender.
Also adds isAttractedTo functionality to all GameCharacters, and allows checks against abstract genders rather than only other characters.


A-Master-of-Puppets:
#357 includes these changes, I'm posting them separately in case #357 is too buggy to accept right now.

CognitiveMist:
Instead of rebuilding the maps every time, you should either use a switch, build the maps outside the functions, or what I would do is move the declaration outside the function, and only build the map if it is null:

```java
Map things = null;
public getThing(Object stuff) {
    if(things == null) {
        things = new Map();
        // construct the map here
    }
    return things.get(stuff);
}
```

CognitiveMist:
@A-Master-of-Puppets 

Innoxia:

Issue #355, opened 2018-04-17
Brothel clients generate as wrong gender


A-Master-of-Puppets:
When selling yourself at the brothel, clients will generate who are not compatible with the PC's orientation (e.g. I'm only into women, and men come to see me).

Innoxia:
Fixed in 0.2.4. :3

Innoxia:
I think. :s 

I'm checking and fixing it now. ^^

PR #357, opened 2018-04-17
Brothel client fixes

### LINKED ISSUE #354: Brothel clients not into PC
When you sell yourself at the brothel, clients not into your gender will generate (e.g. clients only into men when you're a woman).

A-Master-of-Puppets:
This fixes #354 and #355, ensuring that whenever the PC sells themselves in the brothel, they will get a client of the right gender and orientation. May produce some strange results as it currently converts genders directly; for example, it will convert a F_V_B_FEMALE into a M_V_B_BUTCH, which may not be desired. Includes #356 as these methods are needed for it to work.

### DIFF
--- a/src/com/lilithsthrone/game/character/GameCharacter.java
+++ b/src/com/lilithsthrone/game/character/GameCharacter.java
@@ -1980,7 +1980,32 @@ public abstract class GameCharacter implements Serializable, XMLSaving {
 	public void setSexualOrientation(SexualOrientation sexualOrientation) {
 		this.sexualOrientation = sexualOrientation;
 	}
+	
+	public boolean isAttractedTo(GameCharacter character)
+	{
+		return isAttractedTo(character.getGender());
+	}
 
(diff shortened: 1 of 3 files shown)


src/com/lilithsthrone/game/character/GameCharacter.java
```diff
@@ -1980,7 +1980,32 @@ public SexualOrientation getSexualOrientation() {
...
+	}
 
+	public boolean isAttractedTo(Gender gender)
+	{
+		switch(getSexualOrientation())
+		{
+			case AMBIPHILIC:
+				return true;
+			case ANDROPHILIC:
+				if(gender.isFeminine())
```
Pimvgd: can be simplified as return !gender.isFeminine()

Innoxia:
This should be fixed in 0.2.4. Thank you for your work, and I'll take a proper look at #356 soon! ^^

PR #359, opened 2018-04-19
BUGFIX: Fix Lilaya Geisha submissive sex

TadUnlikely:
This fixes the bug introduced in commit 7f723ab where Lilaya will immediately bail out of Geisha submissive sex. The simplest solution seemed to be overriding her isAttractedTo() method to always return true.

No new graphical assets are required, this is simply a bugfix.

Change has been tested on 0.2.3.1. Lilaya no longer bails out due to not having the incest fetish.

@Innoxia: I think you'll want to grab this one for sure before the next release.

Fixes issue #360


Innoxia:
Thank you for this! ^^

PR #363, opened 2018-04-19
Quick bugfix to ensure image files loading on non-windows filesystems.

Irbynx:
Bugfix that allows image files to load on non-windows filesystems by using proper slash (/ instead of \) on non-windows OS. Didn't test on windows since I have no access to it.


Irbynx:
Haven't used windows in a while to be honest, so I might have forgotten about that part, although I do recall that my cross-platform code does use / as a separator with no problems in other programs. I erred on the side of caution with Java since I don't know it all that well.

A-Master-of-Puppets:
I don't think it really has anything to do with Java. By the time it's an issue, the strings have already been handed off to system API calls. At that point, it doesn't matter what language the calling code was written in.

Innoxia:
Thank you for bringing this to my attention! I fixed it by changing the "\\" into "/". ^^

Issue #372, opened 2018-04-21
Personality values not importing.


Git-Gud420:
They export fine, but upon importing all the values become average.

TadUnlikely:
Can confirm on 0.2.3.5.

TadUnlikely:
@Innoxia: I found the problem - the save routine isn't writing the personality entries correctly. I have a fix on my personalitySaveFix branch, but I'll need a little bit of time to write a recovery routine for the existing values. I should have it ready within the next hour or so.

Innoxia:
Thank you for the report and fix! This is fixed in 0.2.4. ^^

PR #373, opened 2018-04-22
BUGFIX: Correct save logic to write out personality entries properly.

TadUnlikely:
Corrects issue #372 for personality values not importing.

The actual problem was that the personality entries were not being written out properly. I changed it to write out the actual personality entries.

Note to @Innoxia: It probably wasn't necessary to comment out the line I did. If you were planning on using that entry for something, feel free to keep it (although I might rename it.)

Nope

Tested on 0.2.3.5.


Innoxia:
Thank you very much for your effort! I fixed this on my side, so it should all be working in the latest version. ^^

PR #420, opened 2018-05-02
Fix typos in telepathic communication upgrades

Mach565:
Just fixed 2 mentions of target when it should have been caster.


Innoxia:
Thank you! :3

PR #446, opened 2018-05-06
Fix intoxication percentage logic

### LINKED ISSUE #443: Intoxication percentage
I believe either the numbers on drinks are wrong, or the intoxication level shown in the perk and text, since after drinking two Wolf Whiskeys I was expecting to see 80% and saw 0.8%. I figure this is a thing in the code that should just be a missing *100 or something.

Pimvgd:
Fix #443 by relocating intoxication percentage logic to a single location (GameCharacter), fixing any issues with it there, and then having everything use the GameCharacter's method of calculating intoxication percentage

Tested on ad09d08835661cc5c983016a31025e52f79e4652 which is 0.2.4.5


Innoxia:
Thank you! ^^

PR #447, opened 2018-05-06
Fix #438 by removing strange characters from svgs

### LINKED ISSUE #438: Many issues with text/symbols appearing in in-game icons
**Version** : 0.2.4.5
**From** : Discord
**Date** : 4 May 2018

image

Pimvgd:
Tested on ad09d08835661cc5c983016a31025e52f79e4652 (0.2.4.5)


Innoxia:
Thank you for finding this problem! I have no idea why Inkscape has inserted non-breaking spaces into those files... It's also added a lot of empty and unused <text> elements, as well as some other random pattern-related ones. I've gone through and removed all of these along with the spaces that you identified. Although I'm not merging this (as I did it in my local code, which I'll push tonight), I will credit you for this fix.

Thank you again for finding the cause of this problem! ^^

PR #449, opened 2018-05-06
Remove modifiers during loading if the savefile explicitly mentions them as false

### LINKED ISSUE #427: Cum is always being described as "sticky" even when that modifier has been removed
**Version** : 0.2.4.5
**From** : Discord
**Date** : 3 May 2018

Pimvgd:
This fixes #427 - most of the bug is already described there. Basically, the game used to load modifiers from a type, but then also from the modifiers themselves. This PR alters the loading to only add modifiers if they're not present already, and if the modifiers are present and the savefile says the character doesn't have them, they're removed.

Tested on ad09d08835661cc5c983016a31025e52f79e4652 (0.2.4.5), although I didn't test all the cases.

**As a result, This needs a proper review.**

### DIFF
--- a/src/com/lilithsthrone/game/character/body/Body.java
+++ b/src/com/lilithsthrone/game/character/body/Body.java
@@ -2,6 +2,7 @@ package com.lilithsthrone.game.character.body;
 
 import java.io.Serializable;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


Innoxia:
Thank you! :3

PR #461, opened 2018-05-10
Improvements to custom image loading

DJ4ddi:
### Goal
The goal of these changes it to allow easy addition of custom images for slaves and other generic characters.

This was previously problematic due to the folder naming based on code class (which overlaps for tons of characters and requires knowing wether the NPC is, for example, a submission or dominion alleyway attacker).

### Changes
- An internal custom artist was added to avoid having to create an artist .xml. This artist does not show up in the preference list and is always preferred over others if images are available.
- The image loader code was improved to allow nearly arbitrary naming (clothed, partial and naked are now only checked as prefixes, not numbered names).
- Images for unique characters are loaded in the same way as before (with the exception that empty artwork lists are culled).
- Images for generic characters are loaded from a seperate directory after an in-game name is available.
- The image modding guide has been updated accordingly.

### Testing
All changes were tested extensively in version 0.2.4.5 and quickly after porting to 0.2.5 and should be safe to merge.

### Caveats and future work
There are a few things that aren't exactely the way I would like them to be, but are hard to fix without lots of additional work.
- Assigning random images to random encounters: This would add lots of possibilities, but provide very inaccurate results in its current state (since characters and images mostly wouldn't match). In order for this to make sense, every image would need tags describing body parts, colors, clothing, etc.
- Automatic renaming of the folder when a character is renamed: A useful quality of life improvement, but currently impractical due to a high duplicate chance for character names. This would interfere with the player's intention (e.g. you rename one character and the images for another move as well). It would be worth adding once characters have persistent unique human readable identifiers (or at least a high chance of a unique name). Until then, it is acceptable to require the player to rename folders.
- Persistent naming after femininity changes: I considered using the name triplet information instead of just the current name in order to do this but realized it may actually be intended behavior to display a different image based on the characters femininity.
- A button to add custom images from within the game: Frankly, the only reason I didn't implement this is the fact that I don't want to meddle with the WebEngine UI. If somebody wants to make a nice button and point me to the code callback, I will happily implement the logic for it.


DJ4ddi:
### Update

I noticed a significant lag when opening character pages with images (which became more relevant since most characters have images on my testing version), so I decided to fix that.

#### Changes

The latest commit adds an image cache that is currently only used for character images. The default image for every character (should there be one) is loaded and cached by a seperate thread after loading a game. Every other image (such as variations or different artists) is only cached after being accessed (and therefore immediately loaded) once.

The _ImageCache_ can in theory be used for any image (although you might want to make changes to _CachedImage_ to avoid unnecessary data). The methods should be fairly simple to understand:
- **requestCache** queues an image for loading if it is not cached already
- **requestImage** attempts to retrieve an image, but may return null if it is not cached (it will be queued instead)
- **getImage** retrieves an image, but may block the caller until it is loaded

#### Profiling

This change is guaranteed to not cause memory issues due to the usage of soft references, which may be cleaned up by the garbage collector when memory gets tight (the references are then weighted by time since last access).

On my system, the average time to parse the image for the character information screen has reduced from 270ms to <1ms.

Innoxia:
Thank you so much for this! I'll try to read through it and get it merged before the hotfix. ^^

DJ4ddi:
### Update

Recent changes on the master branch caused special cases like Brax and Lilaya to break. The latest commit is a quick fix for that. I might look into this further to find a solution that doesn't have duplicated loading code within their classes (see [Brax](https://github.com/Innoxia/liliths-throne-public/blob/master/src/com/lilithsthrone/game/character/npc/dominion/Brax.java) and [Lilaya](https://github.com/Innoxia/liliths-throne-public/blob/master/src/com/lilithsthrone/game/character/npc/dominion/Lilaya.java)).

DJ4ddi:
### Update

The latest commit has some refactoring to improve the way in which special cases like Lilaya and Brax are dealt with. In order to do this, I had to rebase all commits onto a newer version of the master branch (which is why all commits reappeared with newer dates).

Innoxia:
Sorry it's taken me so long to get to this! I'll try and get it merged tomorrow. :3

DJ4ddi:
No worries, I kept up to date with the blog so I'm aware that you intended to sort out pull requests for a while now. Figured it would take some time (since this PR is complex but in no way crucial).

After this is merged, I want to add an option to display character images in the tooltip (since image loading time is no longer an issue). I didn't do this yet to avoid working on an outdated branch and adding even more to your backlog.

DJ4ddi:
### Update
Commit a2c78e7dddc19a72df1ba39a5e3e76edfb0874cd adds the current character image to the attribute overview tooltip (the one that appears when hovering the race icon). Contrary to what I thought previously, a rebase was not necessary and the changes should still be safe to merge.

#### Changes
The attribute tooltip now contains the currently selected character image. It is fetched using a cache request and does not slow down the display of the tooltip. If the image is not cached, it's queued and the tooltip looks like it did previously.
Additionally, I added an option (which is on by default) to disable this behavior. Therefore the tooltip is identical to the previous version if either the artwork or the thumbnail option are disabled.

#### Testing
The changes were tested on 0.2.7.1, along with all previous changes.

DJ4ddi:
Merge conflict resolved with commit 587db46997b285a8f730df107dcf2114fa5aaf68.

DJ4ddi:
### Update

I'll propably run out of things to polish at some point. But today is not that day.

Commit e0bcd43a5f95137e86e2a36c5fd83512901f491e saves artists/images that the user explicitly selected.

#### Changes

A new section `<artwork>` is added to character XML nodes that have images available at the time of the save (it is omitted if there there are none). It contains two optional nodes `<overrideArtist>` and `<overrideImage>` that both have an `"index"` attribute containing the indicies for artists and images respectively. These are only written when the value is not equal to the default.

On loading, both entries are searched and - if present - parsed to select and cache the image that was displayed at the time of the save. Faulty values are fixed through the already existing index parsing procedures.

#### Testing

The changes were tested on 0.2.8.2, along with all previous changes.

Innoxia:
I'm sorry it's taking me so long to merge this! I'll try to read through it all and merge it before I go to sleep today. ^^

Innoxia:
I've run out of time tonight, but will take a look firs thing tomorrow! :3

Innoxia:
Thank you so much for this! The image loading times are completely gone, and the added functionality of adding custom artwork is really cool! :3

Are you on the LT discord? (If so, I could add you to the developers channel. ^^)

PR #467, opened 2018-05-11
fix softlock caused by opening options menu during examining alexandria in intro

Pimvgd:
Opening character examination in intro causes saved dialogue node
Opening options causes saving current dialogue node

This overwrites the previous dialogue node, making you stuck

Solution; don't save when opening options when in character examination in intro


Innoxia:
Thank you! :3

PR #471, opened 2018-05-12
Create method hasWeaponEquipped in GameCharacter for cleanup

Pimvgd:
mainly to fix a possible null pointer when enchanting offhand weaponry (yes, you can't do this, but the code seems to want to handle it) without a main weapon equipped


Innoxia:
Thank you! ^^

PR #472, opened 2018-05-12
Increase utiltext parse performance

Pimvgd:
### Purpose
Improve performance and fix crashes/infinite loops with npcs being called [npc.name].

### Description
I described this on discord:
parsing works by scanning the entire string, then looking for the first executable section, then doing that and starting over from the beginning.
Lower bound complexity is O(n); it has to check each character to find 0 ['s.
Upper bound is infinite loop, as seen when pc.name is [pc.name].
Pretending there are no infinite loops and commands don't output commands,
Upper bound would be O(n^n)
thus if we feed it large slabs of text we're screwed
but here's what I wonder
provided that the output shouldn't contain commands...
why would you need to parse the things you already parsed?
Thus I will try to see what happens if I get rid of the "and start over from the beginning"
```
Before
Took 598.196056 for input of length 2770447
Took 583.028787 for input of length 2770447
Took 566.70835 for input of length 2770447
Took 426.148428 for input of length 2770447
Took 514.132226 for input of length 2770447

After
Took 381.883276 for input of length 2770447
Took 262.659686 for input of length 2770447
Took 296.367686 for input of length 2770447
Took 232.573062 for input of length 2770447
Took 198.476419 for input of length 2770447
```
This is about opening the fetishes screen
and it's in milliseconds

Next, of course, I wonder if the command output really needs to be parsed again
And the answer seems to be "no", although this has no real visible effect in parse times(edited)
It should, however, stop people from creating infinite loops.
meaning you can now call people [npc.name] without the game exploding
I think we should have this in a beta release or something; the non-exe release
I mean, I'm absolutely sure that "not parsing things again you've already found to not contain any commands", that's correct
but "not parsing the result of commands" might not be
The main thing I'm worried of is things like [style.colourPink(feminine)] style things
like, one insertion has occured, but it still needed coloring

### Graphical assets required
None

### Tested?
Yes, on 0261d14169c96f2060f04daae627cb84cf5e12f8 (0.2.5.5 pre-release)

### DIFF
--- a/src/com/lilithsthrone/game/dialogue/utils/UtilText.java
+++ b/src/com/lilithsthrone/game/dialogue/utils/UtilText.java
@@ -4568,73 +4568,65 @@ public class UtilText {
 	private static String parseSyntaxNew(String target, String command, String arguments, List<GameCharacter> specialNPCList) {
 		
 		UtilText.specialNPCList = specialNPCList;
-		parseCapitalise=false;
-		parseAddPronoun=false;
+		parseCapitalise = false;
+		parseAddPronoun = false;
 		
 		if(command.split("_").length==2) {
 			if(Character.isUpperCase(command.split("_")[0].charAt(0)))
 				parseCapitalise = true;
 			command = command.split("_")[1];
-			if(Character.isUpperCase(command.charAt(0)))
-				parseCapitalise = true;
 			parseAddPronoun = true;
+		} 
 			
-		} else {
-			if(Character.isUpperCase(command.charAt(0)))
-				parseCapitalise = true;
+		if(Character.isUpperCase(command.charAt(0))) {
+			parseCapitalise = true;
 		}
 		
-		String output = "";
-		boolean commandFound=false;
-		
-		boolean characterFound = false;
-		targetLoop:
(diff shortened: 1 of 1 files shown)


Innoxia:
I've just spent a while testing this, and it seems to be working fine within the boundaries of what would be expected from normal use. The only thing that I've found which doesn't work is when you nest more than two commands with arguments.

i.e.
`[style.bold([style.colourPink(:3)])]` works correctly, and outputs a bold, pink ":3"
`[style.bold([style.colourPink(:3[style.italics(^^)])])]` on the other hand, outputs a bold, pink ":3[style.italics(^^)]"

I think that this is acceptable behaviour, as I can't imagine a case where you'd need to nest any commands with arguments even once (which works), let alone twice (which doesn't).

The performance gains from this PR are very, very good, especially in any screen where lots of svgs need to be rendered (such as in the inventory screen). Where a full inventory used to have a noticeable lag between clicking on items, when I've tested it on my system just now, it's instant.

I think the behaviour of multiple-nested commands not working is a small price to pay for the huge increase in performance, and am willing to merge this now, unless there are any other concerns you had.

PR #475, opened 2018-05-13
Re-allow parsing nested commands; fixes issues with Lilaya's Lab scene

Pimvgd:
Should not negatively affect performance too much

https://cdn.discordapp.com/attachments/337330311694254121/445193814454566922/unknown.png

" [lilaya.speech([pc.Name]! E-Erm, w-we were just... Doing an experiment! Yes! Isn't that right, Rose?!)]"

Seems we were wrong in #472 that there were no nested commands


Innoxia:
Thanks! ^^

PR #482, opened 2018-05-14
Convert UtilText.parse from recursive to iterative

Pimvgd:
In an attempt to reduce memory footprint. Attempt to properly fix nested command parsing, mainly by checking how many braces the input has. If it has one set of braces, the output is not parsed again. This means no more [npc.name] outputting [npc.name] shenanigans.

It also seems to improve performance for the fetishes screen, although this is slightly less noticeable than before. What it should mostly do is reduce the amount of memory needed.

Tested on 9f05f4642f520cba8db7a4b141e70ddc7fd17b3d (0.2.5.2 release)

### DIFF
--- a/src/com/lilithsthrone/game/dialogue/utils/UtilText.java
+++ b/src/com/lilithsthrone/game/dialogue/utils/UtilText.java
@@ -561,6 +561,12 @@ public class UtilText {
 		return parse(Util.newArrayListOfValues(specialNPC1, specialNPC2), input);
 	}
 	
+	enum ParseMode {
+		UNKNOWN,
+		CONDITIONAL,
+		REGULAR
+	}
+	
 	/**
 	 * Parses supplied text.
 	 */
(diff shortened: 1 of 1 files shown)


src/com/lilithsthrone/game/dialogue/utils/UtilText.java
```diff
@@ -723,29 +713,67 @@ public static String parse(List<GameCharacter> specialNPC, String input) {
...
+			return resultBuilder.toString();
 		} catch(Exception ex) {
 			System.err.println("Failed to parse: "+input);
 			ex.printStackTrace();
 			return "";
 		}
 	}
 	
+	private static boolean substringMatchesInReverseAtIndex(String haystack, String needle, int index) {
+		index++;//this fixes my off by one error and I'm too tired to figure out why
```
Innoxia: W-Well, so long as it works... BlobSweats

Innoxia:
Thank you! :3

PR #484, opened 2018-05-15
Rental Mommy Punctuation Fixes

WoefulWombat:
Just some quick dialogue punctuation corrections for the Mother's Day event.

Remember, a dialogue tag coming after the dialogue itself begins in lowercase, even if the dialogue ended on a question mark, exclamation mark or ellipsis.


Innoxia:
Thank you once again! ^^

I think I understand now. I've been using lowercase as you instruct in the more recent Slime Queen's dialogue. ^^

Issue #505, opened 2018-05-19
Brothel clients generate as wrong gender


A-Master-of-Puppets:
Please forgive the duplicate issue, but for some reason I can't reopen #355. PR #357 (since closed) had fixes for #355 and #354, but only #354 has been fixed upstream.

PR #508, opened 2018-05-19
Brothel client fixes

A-Master-of-Puppets:
Fix for #505, superior to fix from #357; now you will only see things like M_V_B_BUTCH if that is enabled in player gender preferences.


CognitiveMist:
@A-Master-of-Puppets 

I think this would be a great change to add to the game.

However, this PR is out of date with the code it is based upon. Do you still believe it is useful? Do you think it might be worth updating?

PR #520, opened 2018-05-21
Only show horns which are real as selectable for transformation

Pimvgd:
Fixes HornType.NONE showing up as an option for enchanting. Horn removal can be done by selecting the remove option.


IIAkiraII:
unknown
unknown 1

Innoxia:
Thank you! ^^

Issue #519, opened 2018-05-21
Incubi transformations not showing during the following sex scenes


Nenoxx:
**Version** : 0.2.5.2
**From** : Discord
**Date** : 21 May 2018

![image](https://user-images.githubusercontent.com/23719598/40313580-7d5cda02-5d16-11e8-8b24-09c74316f184.png)

**EDIT : Closed, duplicate of** #490

A-Master-of-Puppets:
This looks like a duplicate of #490.

A-Master-of-Puppets:
It was actually a problem if _any_ character tried to transform _any_ incubus like that. I guess nobody noticed it before because incubi are rarer than succubi, and perhaps with the full transform menu now available most people use that instead of the quick one.

Issue #490, opened 2018-05-16
Alley incubi quick TF bugged?


ThatWhichLurks:
I've had an issue where the quick TF options after defeating an incubus in the alleys don't actually change his genitals.  It doesn't affect succubi or the full TF menu.  I've ruled out any saved variables from previous encounters with the same demon being the source of the issue, by reproducing it on a freshly spawned one.  However, the file I discovered it on was a fairly high level PC and I can't rule out more general saved variables glitching things up.

ThatWhichLurks:
UPDATE: Just reproduced it on a pretty fresh character built to set as few flags and variables as possible while being able to clobber the alley demons.  (He's even a virgin!)  Save right before attacking an incubus attached
[test.txt](https://github.com/Innoxia/liliths-throne-public/files/2011638/test.txt)

ThatWhichLurks:
Since the other one about this bug was closed and a fix was put up, marking closed.

PR #521, opened 2018-05-22
Fix Incubus ignoring genital feminisation choices

A-Master-of-Puppets:
Fix for #490 and #519. The choices for "No dick" and "Vagina" didn't actually do anything, presumably a relic of when the encounter was _just_ succubi, and they would have female genitals to begin with.


Innoxia:
Thank you! ^^

PR #524, opened 2018-05-22
Fixed incorrect pronoun

Mach565:
Corrects a mention of 'she' instead of '[npc.she]'.


Mach565:
Looks like Skitt's law is in effect here. Correct a mistake, make one myself.

Innoxia:
Thank you!

PR #652, opened 2018-05-27
Mods added to gitignore

Pimvgd:
you can add files manually with git add -f, but I don't want to see my mods in my unchanged files list because this makes it hard to see what I changed


Innoxia:
Thanks! :3

PR #664, opened 2018-05-30
Add configuration of player character pronouns

GalacticOtter:
This change adds an option for players to chose the pronouns (i.e. first, second or third person) used for their player character. This change only contains the framework (including parser changes), content will be adapted in future work.

The following commands have been added to the parser:

- `pc.youName`: Outputs either a subjective pronoun in first/second person or the PC name in third person.
  Example: I/you/Jane
- `pc.youObjName`: Outputs either an objective pronoun in first/second person or and the PC name in third person.
  Example: me/you/Jane
- `pc.yourName`: Outputs either a dependent possessive pronoun in first/second person or the PC name in third person.
  Example: my/your/Jane's
- `pc.yoursName`: Outputs either an independent possessive pronoun in first/second person or the PC name in third person.
  Example: mine/yours/Jane's
- `pc.inflect(root,first,second,third)`: Inflects a root depending on the pronoun settings.
  Example: `inflect(ha,ve,ve,s)` outputs "have", "have", and "has" in first, second, and third person respectively.
- `pc.inflectS(root)`: Inflects a root for the common case where the third person singular has a suffixed "s".
  Equivalent to `inflect(root,,,s)`.

The commands `you`, `youObj`, `your`, `yours`, and `yourself` have been added as aliases to `heShe`, `himHer`, `hisHer`, `hisHers`, and `himself`/`herself` respectively.

-----

- No new graphical assets are required.
- This change has been tested on version 0.2.5.8.
- I am Otter#4718 on discord.

### DIFF
--- a/src/com/lilithsthrone/game/Properties.java
+++ b/src/com/lilithsthrone/game/Properties.java
@@ -44,6 +44,7 @@ import com.lilithsthrone.game.settings.ForcedFetishTendency;
 import com.lilithsthrone.game.settings.ForcedTFTendency;
 import com.lilithsthrone.game.settings.KeyCodeWithModifiers;
 import com.lilithsthrone.game.settings.KeyboardAction;
+import com.lilithsthrone.game.settings.PlayerPronouns;
 import com.lilithsthrone.main.Main;
 
 /**
@@ -88,7 +89,8 @@ public class Properties implements Serializable {
 	public Map<KeyboardAction, KeyCodeWithModifiers> hotkeyMapPrimary, hotkeyMapSecondary;
 
 	public Map<GenderNames, String> genderNameFemale, genderNameMale, genderNameNeutral;
-	public Map<GenderPronoun, String> genderPronounFemale, genderPronounMale;
+	public Map<GenderPronoun, String> genderPronounFirst, genderPronounSecond, genderPronounFemale, genderPronounMale;
+	public PlayerPronouns playerPronouns = PlayerPronouns.SECOND_PERSON;
 	
 	public Map<Gender, Integer> genderPreferencesMap;
 	private Map<Subspecies, FurryPreference> subspeciesFeminineFurryPreferencesMap, subspeciesMasculineFurryPreferencesMap;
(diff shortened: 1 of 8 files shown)


TadUnlikely:
The majority of this pull request looks okay, although I haven't worked out yet if it really simplifies how the dialogue works sufficiently.

I do note that there are some problems with the "inflect" and "inflectS" parser commands. Asking the dialog writer to supply your verb tenses is clunky. Also, there are a limited number of rules for verb tenses in English, and they can be encoded so all the dialog writer has to think is "What verb do I want here?"  See my new "verb" parser from #706 for how this should probably work.

TadUnlikely:
I've had a couple of days to think about it, and the "inflect" command might has two advantages that might make it a useful fallback.

The first is allowing for non-standard verbs. I haven't been able to come up with any that aren't covered by the rules in the "verb" command or the sheIs/sheHas logic,  but it's possible as the language advances.

Probably the more useful reason for having it would be easier internationalization. This is okay, although it might make more sense in that case to build a more generalized "verb" command that uses maps for the varying verb tenses amongst differing languages. That would probably be a lot of work, though, and I honestly don't know enough languages to know if it would really work.

GalacticOtter:
@TadUnlikely You're right, the `verb` command you've implemented does seem more appropriate. I'll fix the conflicts in this PR and update it to use `verb`.

GalacticOtter:
PR updated and conflicts fixed.

Innoxia:
I feel as though refactoring all of the game's dialogue to be suitable for this change would be a huge task, but it's more the fact that all of the content that I write in the future would need to follow this new formatting that makes me unwilling to merge this.

I really do appreciate the amount of effort you've put into this, but I just can't justify the increase in writing time/effort that would be a result of merging this PR. For an example of what I mean, take this example sentence:
`Both you and your companion are taken off guard, and, although you try to struggle out of their grip, the enforcers quickly overpower the two of you.`
Would need to be written as:
`Both [pc.youName] and [pc.yourName] companion are taken off guard, and, although [pc.youName] [pc.verb(try)] to struggle out of their grip, the enforcers quickly overpower the two of [Missing:(us/you/them)].`

Not only is this a lot more cumbersome to type, but it also makes proof-reading considerably more challenging. I am also concerned with the way structures such as `Both you and your companion` are impacted. In first-person, I believe this would be grammatically inaccurate as `Both I and my companion...`, and should instead be `Both my companion and I...`

In conclusion, I do appreciate the thought and effort put into this, but for a feature that has maybe been requested once or twice, I simply cannot justify the increase in writing workload. Sorry. :(

PR #675, opened 2018-06-01
Util method to select random item from list

A-Master-of-Puppets:
Seeing the pattern `list.get(Util.random.nextInt(list.size()))` all over the place is driving me crazy. Here's a method added under Util so you don't have to keep repeating that. With method overloading, you could also swap out certain lists for weighted maps if you want without having to refactor thousands of lines of code. The downside is that there _are_ thousands of lines of code that will need replacing to fully implement this; I'm nowhere near fixing all of them. Having this available for new code/refactoring still seems like a good thing even with old instances of the pattern in place, however.


Innoxia:
Thank you! I'll use this from now on, and will refactor any of the old `list.get(Util.random.nextInt(list.size()))` patterns I come across. ^^

PR #692, opened 2018-06-03
Fix encyclopedia weapon colours

### LINKED ISSUE #668: The colour preview for discovered weapons doesn't work.
**Version** : 0.2.5.8
**Date** : 31 May 2018

image
image

Pimvgd:
I'm not sure that passing null for the primary and secondary colours is correct, but I can see the different colours in encyclopedia now. Fixes #668.


Innoxia:
There shouldn't be any problem with passing in null. The method that `getSVGImage()` calls checks for nulls, so it's all good. :3

PR #693, opened 2018-06-03
Fix history status effect not properly applied

### LINKED ISSUE #666: Newly made characters have mysterious -5 penalty to lust resistance
Myself and one other person on the Discord have been able to produce characters that have a -5 penalty to lust resistance, applied on top of the normal "visible" bonuses you get.

A save file that the other person has provided includes these status effects.

```
<statusEffect type="PURE_HUMAN" value="-1"/>
<statusEffect type="PHYSIQUE_PERK_1" value="-1"/>
<statusEffect type="INTELLIGENCE_PERK_3" value="-1"/>
<statusEffect type="CORRUPTION_PERK_0" value="-1"/>
<statusEffect type="AROUSAL_PERK_0" value="-1"/>
<statusEffect type="LUST_PERK_1" value="-1"/>
<statusEffect type="WEATHER_STORM_PROTECTED" value="-1"/>
<statusEffect type="ORIENTATION_AMBIPHILIC" value="-1"/>
```

Only two of these appear to adjust lust resist, and they both increase it, rather than decrease it. Lust resist value for this character should be 0 base and +45 temporary, but it's 0 base and +40 temporary for whatever reason.

Pimvgd:
Fixes #666 , Player was given UNEMPLOYED which raised lust resist by 5, then the bonus attribute map is reset, then the history is set to what it should be, which "reverses" the effects, leading to -5 lust resist.

### DIFF
--- a/src/com/lilithsthrone/game/character/GameCharacter.java
+++ b/src/com/lilithsthrone/game/character/GameCharacter.java
@@ -377,8 +377,6 @@ public abstract class GameCharacter implements Serializable, XMLSaving {
 		this.setLocation(Main.game.getWorlds().get(worldLocation).getCell(startingPlace).getLocation());
 		homeLocation = location;
 		
-		history = History.UNEMPLOYED;
-		
 		// Set up personality:
 		personality = new HashMap<>();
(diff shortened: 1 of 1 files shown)


Innoxia:
Ah, nice find! :3

PR #697, opened 2018-06-03
Fix for crash with bimbo character dirty talk

A-Master-of-Puppets:
This removes extraneous spaces from the bimbo-talk word list, which could cause empty strings to be passed to isEndOfSentence if further modified with muffled or sex sounds. Also changes isEndOfSentence to handle being passed empty strings, as defence-in-depth against possible future double-space issues.


Innoxia:
Thank you!

PR #358, opened 2018-04-18
Dialogue Improvements

TadUnlikely:
Improves some existing dialogue parsers and adds others. These parser changes allow for more easily combined dialogs. I've included a few sample changes to GenericOrgams to demonstrate both usage and how it can reduce code bloat (and likely bugs) by allowing for the combination of PC-specific dialogues with the matching NPC dialogues.

Especially notable are the handling of verb persons (you reach, she reaches) and the addition of namePos (which eliminates the need to add 's to possessive nouns.)

I wrote this sometime back, and actually did an almost full overhaul of GenericOrgasm. I can do another pull request after this is accepted to clean up the rest of the file, but I'll want to be careful about merging my changes, given how that file has changed since I wrote the code. 

Tested on 0.2.3.1, no new graphical assets.

PR #706, opened 2018-06-04
Dialogue improvements (New Parser Options)

TadUnlikely:
Adds some parser commands that will simplify creating dialogue. GenericOrgasms.java is full of if/else statements where the only difference between the player and NPC dialogue is the person of the verbs and possibly second/third person pronouns. This accounts for both of those options, and also includes a sample of the kind of code reduction this pull request will allow. I can do a series of follow-up commits to take full advantage of these new features, but will want this merged before I start that work. This was originally #358, but it sat unmerged so long that it became impossible to merge due to changes in UtilText.

Improves some existing dialogue parsers and adds others. These parser changes allow for more easily combined dialogs. I've included a few sample changes to GenericOrgams to demonstrate both usage and how it can reduce code bloat (and likely bugs) by allowing for the combination of PC-specific dialogues with the matching NPC dialogues.
No new graphical assets

Conceptualized previously, this has been updated and tested most recently on 0.2.6.2.

N/A

### DIFF
--- a/src/com/lilithsthrone/game/dialogue/utils/UtilText.java
+++ b/src/com/lilithsthrone/game/dialogue/utils/UtilText.java
@@ -941,6 +941,13 @@ public class UtilText {
 				+ " If a prefix is provided, the prefix will be appended (with an automatic addition of a space) to non-capitalised names."){
 			@Override
 			public String parse(String command, String arguments, String target) {
+				if(character.isPlayer()) {
+					if(command.startsWith("N")) {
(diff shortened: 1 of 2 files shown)


Innoxia:
Sorry that it took me so long to get around to this! Your new parsing options will be extremely useful for the sex action work I'll be doing in 0.2.7! Thank you so much for this!

I'm going to start reading through it properly now, so it should be merged within the hour. :3

PR #715, opened 2018-06-04
Fix addiction timer displays to show 22:00 instead of 21:60

Pimvgd:
Doesn't fix any issue mentioned in issues but I spotted it myself soooooo why not?


Innoxia:
Thank you! :3

PR #717, opened 2018-06-05
Punctuation Fixes for Slime Queen Quest

WoefulWombat:
We're getting there.

Dialogue punctuation corrections for the Slime Queen quest and some other stuff around Submission. I'll look into the Gambling Den another time.
I'm sure I missed some, and there're the usual debates about whether something should be considered a dialogue tag or not. In which case it's mostly up to preference.

I don't know why it keeps creating an extra empty line after `</dialogue> `at the end of the .txt files and I can't seem to make it stop doing that.


Innoxia:
Thank you once again! ^^

I think I've finally got how it works now. ^^

PR #735, opened 2018-06-06
Update Perk.java

Rfpnj:

Adjusts wording for Nymphomaniac perk
^
no
no


Innoxia:
I actually intended this perk to double essence gains from every orgasm, so the description is correct. I need to take a look at why someone reported this as not working...

PR #738, opened 2018-06-07
Fix self breast milking always emptying breasts

Pimvgd:
And fixed Milk Self Girlcum option saying Milk Self Cum when energy too low


Innoxia:
Thank you!

Issue #748, opened 2018-06-08
When submitting to Epona after trading in six stamps, a name is missing from one of the paragraphs


Nenoxx:
**Version** : 0.2.6.5
**Date** : 9 June 2018

Reporting this issue from Patreon

*"If it's ok with you, could you take charge? you ask, stepping around behind the counter before moving up next to Epona.
INVALID_TARGET_NAME the horse-girl says, placing a 'Be back soon' sign on the counter"*

WoefulWombat:
PR #749 should fix that.

Innoxia:
Fixed by WoefulWombat's PR. :3

PR #749, opened 2018-06-09
Punctuation Fixes for the Gambling Den and Issue #748

WoefulWombat:
Dialogue punctuation corrections for the Gambling Den and a quick fix for #748.

Here, I'm conflicted about whether something like _"[...] turns around to address the room"_ should be a dialogue tag or not. I think it should be, and it does feel right, but I can't point at any rules for this one. Change the _"to"_ to an _"and"_ and it clearly is a tag. But couldn't find anything about it, so I went with my gut.


Innoxia:
Thank you so much yet again! <3

Innoxia:
I also think that "_[...] turns around to address the room_" should be a dialogue tag. :3

PR #755, opened 2018-06-10
Add cum regeneration mechanics toggle

Pimvgd:
I've tested this to see if it works, but I'm not sure I missed anything. Would recommend merging before tuesdays so others can test as well.


Innoxia:
I've read through your PR, and can't see anything that you've missed. Thank you for this! :3

PR #757, opened 2018-06-11
Fix invalid HTML

ltpatch:
- `</br>` should be `<br>` or `<br/>`
- `<hr></hr>` should be `<hr>` or `<hr/>`


Innoxia:
Thank you for this! There are some merge conflicts with the most recent version, so I'll do a search and replace on my end. I'll credit you for this fix in the patch notes. ^^

PR #763, opened 2018-06-12
Add Hair-Type to Slime Transformations in BodyChanging.java

HarelMym:
Adds the Hair Type change to the Slime Transformations, allowing a character to be changed to a 'Greater' whatever.

This permits greater Angels without cheating, as well as Greater Morphs via the Biojuice Canister and a Slime Elixer (Arcane Boosted to Flesh).

Compiled and tested in-game (2.6.5); successful for Angel, Demon, Imp, Reindeer, Canine, Equine, Alligator and Wolf Morphs (Sloths, Orangutans, Breakfast Cereals...) all the -Morphs work with this.


Innoxia:
Thank you for this! :3

I've implemented this in the dev branch, so while I won't need to merge this, it will make it into the next update. I'll credit you for this in the patch notes. ^^

PR #766, opened 2018-06-12
Add configuration of NPCs sexual orientation encounter rates

GalacticOtter:
This change adds an option for players to set the encounter rates for the sexual orientation of random NPCs, in the same way as gender encounter rates.

The generation function uses both the generic player-defined rates and call-specific rates when computing the final orientation. The specific rates can be used for race/feminity-dependent rates. For example, harpies were previously hardcoded as being almost always gynephilic, and now use the generation function with unbalanced specific rates.

-----

- No new graphical assets are required.
- This change has been tested on version 0.2.6.5.
- I am Otter#4718 on discord.

### DIFF
--- a/src/com/lilithsthrone/game/Properties.java
+++ b/src/com/lilithsthrone/game/Properties.java
@@ -28,6 +28,7 @@ import com.lilithsthrone.game.character.gender.AndrogynousIdentification;
 import com.lilithsthrone.game.character.gender.Gender;
 import com.lilithsthrone.game.character.gender.GenderNames;
 import com.lilithsthrone.game.character.gender.GenderPronoun;
+import com.lilithsthrone.game.character.persona.SexualOrientation;
 import com.lilithsthrone.game.character.race.FurryPreference;
 import com.lilithsthrone.game.character.race.Race;
 import com.lilithsthrone.game.character.race.Subspecies;
@@ -91,6 +92,7 @@ public class Properties implements Serializable {
 	public Map<GenderPronoun, String> genderPronounFemale, genderPronounMale;
 	
 	public Map<Gender, Integer> genderPreferencesMap;
+	public Map<SexualOrientation, Integer> orientationPreferencesMap;
 	private Map<Subspecies, FurryPreference> subspeciesFeminineFurryPreferencesMap, subspeciesMasculineFurryPreferencesMap;
 	private Map<Subspecies, SubspeciesPreference> subspeciesFemininePreferencesMap, subspeciesMasculinePreferencesMap;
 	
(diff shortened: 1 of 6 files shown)


GalacticOtter:
Hey @Innoxia, would it be possible to get a review of this PR, or to get it merged?

Innoxia:
Sorry it's taken me so long to respond! I'll try and read through this and get it merged before I go to bed tonight. ^^

Innoxia:
I've run out of time tonight, but will take a look firs thing tomorrow! :3

src/com/lilithsthrone/game/character/persona/SexualOrientationPreference.java
```diff
@@ -0,0 +1,62 @@
...
+		int gyne = Main.getProperties().orientationPreferencesMap.get(SexualOrientation.GYNEPHILIC) * gynephilicWeight;
+		int ambi = Main.getProperties().orientationPreferencesMap.get(SexualOrientation.AMBIPHILIC) * ambiphilicWeight;
+		int andro = Main.getProperties().orientationPreferencesMap.get(SexualOrientation.ANDROPHILIC) * androphilicWeight;
+		int total = gyne + ambi + andro;
+
+		if(total == 0) {
+			return SexualOrientation.AMBIPHILIC;
+		}
+		
+		int random = Util.random.nextInt(total)+1;
```
picobyte: would this also work instead of the newTotal variable? 
```
if (random <= gyne) {
	return SexualOrientation.GYNEPHILIC;
}
if (random <= gyne + ambi) {
	return SexualOrientation.AMBIPHILIC;
}
return SexualOrientation.ANDROPHILIC;
```

src/com/lilithsthrone/game/dialogue/utils/OptionsDialogue.java
```diff
@@ -1191,6 +1196,98 @@ private static String getGenerPreferencesPanel(PronounType type) {
...
+		public String getContent(){
+			return "";
+		}
+		
+		@Override
+		public Response getResponse(int responseTab, int index) {
+			 if (index == 0) {
+				return new Response("Back", "Go back to the options menu.", OPTIONS);
+				
+			}else {
```
picobyte: can simply return null; (no else branch necessary - the if branch also returns)

CognitiveMist: There are many ways to express this code, and the decision should be made with readability in mind. (Which, of course, is just opinion. My opinion is either leave as-is, or use a conditional expression across three lines as below.)

```
return index == 0
	? new Response(...)
	: null;
```

src/com/lilithsthrone/game/Properties.java
```diff
@@ -685,6 +708,21 @@ public void loadPropertiesFromXML(){
...
+				if(element!=null && element.getElementsByTagName("preference")!=null) {
+					for(int i=0; i<element.getElementsByTagName("preference").getLength(); i++){
+						Element e = ((Element)element.getElementsByTagName("preference").item(i));
+						
+						try {
+							if(!e.getAttribute("orientation").isEmpty()) {
+								orientationPreferencesMap.put(SexualOrientation.valueOf(e.getAttribute("orientation")), Integer.valueOf(e.getAttribute("value")));
+							}
+						} catch(IllegalArgumentException ex){
+						}
```
picobyte: shouldn't there be a warning here?

CognitiveMist: There's lots of empty catch blocks in LT. Ideally, each catch block should report an error, or include a comment that describes why it's ok to ignore the error.

Innoxia:
Thank you for this! Sorry it took me so long to merge. ^^

PR #791, opened 2018-06-18
Clothing Enchantments alter Vaginal Urethra stats

HarelMym:
Clothing enchantments can't currently alter vaginal urethral stats; there's no way in the UI to do so, and even 'cheated' in via save editing, they don't show up in the items and are removed if put in. With these changes to AbstractItemEffectTypes, they can be enchanted in clothing, they show up in UI and they function.

I've tested this in build 0.2.7 on the PC, and on several Enslaved NPCs, and it works; it allows altering the urethral stats via clothing enchanted by the PC.

@Innoxia this is obviously just how I assume that the game should work from the rest of the available enchantments; if the current paradigm is intentional it'd be nice to know.


Innoxia:
Thank you! <3

PR #804, opened 2018-06-19
Small note on JDK Compatibility in the README

r79:
Just a very very small PR to help other devs / people who like building it on their own save some time.

There might be more to come ;)


ltpatch:
It works with OpenJFX.

Innoxia:
Thank you! :3

PR #814, opened 2018-06-22
Parameterize and deduplicate code for offspring encounters.

uglybead:

Cleanup offspring encounters so that it is straightforward to add items to affect their encounter rate.

There should be no changes to the effect of the code. Duplicate code has been moved into helper functions.

No.

Yes. Tested against public master.

n/a


Innoxia:
Thank you! :3

Issue #823, opened 2018-06-28
Exception when trying to change position when having sex with a slave in the stocks


Nenoxx:
**Version** : 0.2.7.8
**Date** : 28 June 2018

Error.log after trying to select "Perform oral" : 

`BlockedParts loading failed. Code 1 elbow-length latex glove
AbstractClothingType loading failed. Cause: 'blockedPartsList' element unable to be parsed.
java.lang.NullPointerException
	at com.lilithsthrone.game.sex.Sex.updateAvailableActions(Sex.java:3311)
	at com.lilithsthrone.game.sex.Sex.setSexManager(Sex.java:3271)
	at com.lilithsthrone.game.sex.sexActions.universal.StocksSex$2.applyEffects(StocksSex.java:113)
	at com.lilithsthrone.game.sex.sexActions.SexActionInterface.baseEffects(SexActionInterface.java:186)
	at com.lilithsthrone.game.sex.Sex.endSexTurn(Sex.java:1230)
	at com.lilithsthrone.game.sex.sexActions.SexActionInterface$1.effects(SexActionInterface.java:580)
	at com.lilithsthrone.game.dialogue.responses.Response.applyEffects(Response.java:184)
	at com.lilithsthrone.game.Game.setContent(Game.java:1512)
	at com.lilithsthrone.controller.eventListeners.SetContentEventListener.handleEvent(SetContentEventListener.java:36)
	at javafx.web/com.sun.webkit.dom.EventListenerImpl.fwkHandleEvent(Unknown Source)
	at javafx.web/com.sun.webkit.WebPage.twkProcessMouseEvent(Native Method)
	at javafx.web/com.sun.webkit.WebPage.dispatchMouseEvent(Unknown Source)
	at javafx.web/javafx.scene.web.WebView.processMouseEvent(Unknown Source)
	at javafx.web/javafx.scene.web.WebView.lambda$registerEventHandlers$3(Unknown Source)
	at javafx.base/com.sun.javafx.event.CompositeEventHandler$NormalEventHandlerRecord.handleBubblingEvent(Unknown Source)
	at javafx.base/com.sun.javafx.event.CompositeEventHandler.dispatchBubblingEvent(Unknown Source)
	at javafx.base/com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(Unknown Source)
	at javafx.base/com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(Unknown Source)
[...]

Issue #815, opened 2018-06-22
Issues with the slave stocks : Game locks when preparing to orgasm and can't change position


Nenoxx:
**Version** : 0.2.7.5
**Date** : 22 June 2018

*"Seems to be quite a few bugs with slave stocks in Slaver Alley.  I cannot change position.  If I get to the point of orgasming, all buttons disappear and the game locks.  Also the slaves seem to have free use of their hands ."*

Save and error.log : 
[AutoSave_and_errorlog.zip](https://github.com/Innoxia/liliths-throne-public/files/2128471/AutoSave_and_errorlog.zip)

rantingpirate:
Found ~~the~~ a source of ~~the~~ a problem. When I try to [Move to front] in the stocks scene, Java throws a null pointer exception from line 3287, column 132 of src/com/lilithsthrone/game/sex/Sex.java, where `Sex.getSexPositionSlot(target)` is returning `NULL`, which freaks out the `.get()` that it's wrapped in. From what I can tell, this is because it's looking in the MILKING_STALL_LOCKED_IN_MILKING_STALL table for STOCKS_RECEIVING_ORAL. Not sure why it's looking there though...

I also noticed similar odd and game-breaking things happening in Lilaya's 'Tests' scene. In addition to the disappearing buttons, she did nothing but change position.

Nenoxx:
`I also noticed similar odd and game-breaking things happening in Lilaya's 'Tests' scene. In addition to the disappearing buttons, she did nothing but change position.`

Yes, this has been fixed in the 0.2.7.6 hotfix

PR #833, opened 2018-06-30
Fix for crash during stocks sex

CognitiveMist:
This should address #815 and #823.

The stocks were trying to use the milking stall interactions in
some areas, and the stocks interactions in others.


Innoxia:
I've fixed this on my end using your code, so I'll credit you in the patch notes. Thank you! :3

PR #836, opened 2018-07-01
Fix parser where none of the capitalise commande would not be recongnized

Lightcanadian:
I have found a bug in the parser where the capitalized technique of making the first letter capital would not work even for the given example in the help. It seems that when removing the name from the command it would leave some empty char
[brax.Height] -> 'n/a'n/a'n/a'n/a'H'e'i'g'h't'

~~Now it looks for the first letter~~

An `if` statement was always returning true. So I fix the `if` so that now it properly filter the letter and most importantly remove the character `\U200B` that would appear in the input.

### DIFF
--- a/src/com/lilithsthrone/game/dialogue/utils/UtilText.java
+++ b/src/com/lilithsthrone/game/dialogue/utils/UtilText.java
@@ -680,7 +680,7 @@ public class UtilText {
 					}
 				}
 				
-				if (openBrackets>0 && ((target!=null && command!=null) || String.valueOf(c).matches(".") || c!=' ' || c!=Character.MIN_VALUE)) {
+				if (openBrackets>0 && (Character.isLetterOrDigit(c) || c=='+' || c=='.' || c=='[')) {
 					sb.append(c);
 				}
 				
--- a/src/com/lilithsthrone/game/dialogue/utils/UtilText.java
+++ b/src/com/lilithsthrone/game/dialogue/utils/UtilText.java
@@ -652,7 +652,7 @@ public class UtilText {
 						
 						} else if (c == '(') {
 							if(command == null) {
-								command = sb.toString().substring(1); // Cut off the '.' at the start.
+								command = sb.toString().substring(1); // Cut off the '(' at the start.
 								sb.setLength(0);
 							}
 							
@@ -680,7 +680,7 @@ public class UtilText {
 					}
 				}
 				
-				if (openBrackets>0 && (Character.isLetterOrDigit(c) || c=='+' || c=='.' || c=='[')) {
+				if (openBrackets>0 && (Character.isLetterOrDigit(c) || c=='+' || c=='.' || c=='[' || c=='(')) {
 					sb.append(c);
 				}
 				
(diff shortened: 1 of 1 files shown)


src/com/lilithsthrone/game/dialogue/utils/UtilText.java
```diff
@@ -680,7 +680,7 @@ public static String parse(List<GameCharacter> specialNPC, String input) {
 					}
 				}
 				
-				if (openBrackets>0 && ((target!=null && command!=null) || String.valueOf(c).matches(".") || c!=' ')) {
+				if (openBrackets>0 && ((target!=null && command!=null) || String.valueOf(c).matches(".") || c!=' ' || c!=Character.MIN_VALUE)) {
```
CognitiveMist: This change doesn't seem to work for me locally. If this condition was changed to always exclude non-breaking spaces (`openBrackets>0 && c!='\u200b'`) then it does work for me. Hopefully this isn't a java implementation thing.

As a side note, `c!=' '` is true for all characters except a space. `String.valueOf(c).matches(".")` is true for all characters (except maybe linebreaks, depends on java implementation), but definitely true for a space. (Maybe this was supposed to be `"\\."`? Which then it could be `c=='.'` instead.)

Regardless, `c!=' ' || c!=Character.MIN_VALUE` is equivalent to always true as well.

Therefore, the second clause is always true, and the only condition that matters in the current code is `openBrackets>0`.

src/com/lilithsthrone/game/dialogue/utils/UtilText.java
```diff
@@ -587,8 +587,8 @@ public static String parse(List<GameCharacter> specialNPC, String input) {
 			
 			int startedParsingSegmentAt = 0;
 			
-			for (int i = 0; i < input.length(); i++) {
-				char c = input.charAt(i);
+			for (int i = 0; i < input.length(); i++) {
```
CognitiveMist: You should try to avoid non-changes like this line whenever possible.

src/com/lilithsthrone/game/dialogue/utils/UtilText.java
Lightcanadian: The Character.MIN_VALUE represent an empty char(or null) because for some reason there's a bunch of them. It's also not a consistent issu sometime there's a bunch of them and sometime it just work's. All its doing is ignoring all the empty char that are present
inked2018-07-01 23_48_49-lilith s throne 0 2 7 8 alpha not responding _li

CognitiveMist: I see this (`'\u200b'` is a zero-width space):
image

src/com/lilithsthrone/game/dialogue/utils/UtilText.java
```diff
@@ -652,7 +652,7 @@ public static String parse(List<GameCharacter> specialNPC, String input) {
 						
 						} else if (c == '(') {
 							if(command == null) {
-								command = sb.toString().substring(1); // Cut off the '.' at the start.
+								command = sb.toString().substring(1); // Cut off the '(' at the start.
```
CognitiveMist: It is actually cutting off a '.' in the most common case, here. The '(' is the end marker of the substring.

Lightcanadian: I thought it was a copypasta mistake

Innoxia:
Thank you for this! I've implemented it along with some other changes in my local code, so I'll close this and credit you in the changelog. ^^

PR #839, opened 2018-07-02
Further implement the conceal system

Lightcanadian:
When the race was concealed, the tile text would still display the race.

After the second commit
There was more to it then I thought. Now when the race is concealed:
-The character present window won't show stats
-The character present window won't show the race or the name of the person
-The icon shown is now a question mark(already implemented)
-The character present dialogue won't show the character detail

The only thing that I think is missing is the stats window in the character present dialogue
2018-07-03 22_47_23-lilith s throne 0 2 7 8 alpha

I'm not sure of what direction to take with this window, remove it? Remove all the information from it?

PS. Right now no generated character or unique character return true at `isRaceConceal()` method you need to edit an npc to test the feature


Innoxia:
Thank you! :3

Innoxia:
I made some slight changes to parts of this, but kept the main improvement (the change when standing on an unknown race's tile) the same. ^^

PR #880, opened 2018-07-24
Craft and use buttons on crafting dialogue.

orth0dox:
Crafting potion that you instantly wanna drink and then searching through inventory to find it is extremely annoying. It would be great to do both things at once. 

Added 2 kind of buttons on crafting dialogue: craft and use on self and craft and use on them. Both appear to potions and clothes only. If you have inventory interaction with npc, you can use crafted items on them.

No
Yes, 2.9.1


CognitiveMist:
@orth0dox 

Innoxia:

PR #881, opened 2018-07-24
Custom image polishing

DJ4ddi:
As mentioned on Discord, here are some minor fixes and polishing for my recent PR (#461).

### Goals

There are two major goals covered here: Improved **caching** behavior and image **scaling**.

The cache previously dropped all images before allocating additional memory which led to somewhat frequent wipes. This happened regardless of how much memory was currently used.

Scaling is an issue because the JFX WebEngine uses a primitive, one-step approach, resulting in terrible quality, especially for significantly oversized images.

### Changes

To prevent cache wipes, a more aggressive caching strategy was implemented that only resizes when a new image is added (rather than before every garbage collection). Control over the desired size is explicit and currently configured for at least 8 entries, at most 1/10 of the maximum memory and dynamic growth in between. Under normal circumstances the cache may store around 50 MB.
Furthermore, a cache request is issued for all characters in the cell whenever the player enters it.

To make the overall cache size more consistent, images are downscaled to at most 600x600 (which is only a few pixels smaller than the highest possible display size). This means that the average size for a single image is now a lot more predictable (500 KB).
A positive side effect of this is more control over the scaling algorithm used: Since the resizing is almost exclusively done in a background thread, a high quality approach downscales over multiple iterations with bilinear filtering, decreasing artifacts in oversized images. This is especially noticable in tooltips.

Other than that, some minor issues were resolved:

- Naked player images are no longer hidden.
- The image selection controls now work for the player.
- Newly created NPCs (e.g. after starting a new game) now have their artwork loaded immediately.
- Unrecognized file name prefixes are now loaded as clothed.
- The tooltip image is now hidden if it is locked in the info screen.
- JPG images actually work

### Testing

All changes were tested on version 0.2.8.2 and 0.2.9.1.


DJ4ddi:
Merge conflict resolved with commit ce90f99ba532da72e8395bac49a1acbb30013005.

DJ4ddi:
### Update

After reading a surprised "You can do that?" from people being told about custom images, I decided to finally add a button that imports images without having to meddle with the folders. That's what commit dcf74c7c6c85467a8dbc09f48b3ee4f78d2e15f7 does. Note that it touches a lot of lines because of indentation and imports, it's actually not as big as it looks.

Looking back at it now, assuming that everyone would just find and read the modding.txt was a little naive.

#### Changes

A small button with a "+" icon is now placed next to the artist information button (or in the same spot, if there is no artwork). Clicking it opens a dialogue that allows selecting one or more image files which are copied into the character's folder and imported into the game.

On most platforms, the same dialogue also allows renaming or deleting images.

#### Testing

The change was tested on version 0.2.9.5, _but only on Windows 10, 64 bit_. This is relevant because the JavaFX File Chooser uses native code and has no guarantees for working the same on all platforms (it should, but I cannot confirm it).

Innoxia:
Really fantastic improvements. The image scaling change has got rid of the jagged outlines on some of the larger images, and the custom artwork addition feature is super good.

Thank you so much for all this!

PR #882, opened 2018-07-26
Key binding presets

DJ4ddi:
### Goal

I personally disliked the default key bindings, especially the response box layout (1 to 5 for the first row, 6 to 9 and Ctrl+1 for the second, Ctrl+2 to Ctrl+5 and 0 for the third) is very unintuitive. It is also impractical for one handed usage, which is the standard use case (since the second hand is either on the mouse or unavailable) due to buttons like numbers >7, I and P.

But since optimal bindings are heavily subjective (which is why the settings are necessary in the first place) simply changing the defaults felt insufficient. Binding presets elegantly solve this issue by providing users with several default variants as well as the option to create and share their own setups.

### Changes

- Keybinding presets can now be saved and loaded from the options menu. They are stored in the /res/keybinds folder.
- The order of the movement keys has been changed to WASD (instead of WDSA)
- The cursor navigation keys were missing from the options menu and have been added. The scrolling keys are still missing because they currently don't do anything (?).
- The defaults for some keys changed to a more intuitive, one handed layout.

### Testing

All changes have been tested on 0.2.9.1.


Innoxia:
This is very cool - thank you for this! ^^

PR #883, opened 2018-07-28
inkscape palette

gumyns:
Adds Inkscape palette with predefined colors for rendering SVG. Helps modding clothes ofc.
Colors are taken from there https://github.com/Innoxia/liliths-throne-public/blob/master/src/com/lilithsthrone/rendering/Pattern.java#L84

No

Yes, with Inkscape...


CognitiveMist:
I don't if the author of this PR is still around (and assume they are not, since their branch is deleted), but changes have happened between then and now. I put together an updated palette a few days ago, it can be found here:

https://discordapp.com/channels/302617912949211136/485212939478106134/712485446797623346

@Innoxia maybe just close this PR?

Innoxia:
Thank you for the PR, gumyns, and I'm sorry that it took me so long to get around to looking at this. As per CognitiveMist's suggestion above, I'm going to close this now.

CognitiveMist:
Gumyns, if you want to bring this back, you have my permission to use my palette file in whatever form you see fit.

Issue #152, opened 2018-03-04
'#' character in save path prevents saving


Innoxia:
Found a Bug with saving game, if there is a '#' anywhere in the file path, (For example: C:\#Flash\Lilith's Throne v0.2.0.2 ) you can't save your game on either the exe or jar versions.
I've also checked pretty much all the other symbols on a standard qwerty keyboard and they all seem to work fine, except '#'. You can still load files, just not overwrite or save make new saves.
(I didn't bother to check other various combinations of the symbols besides '()' '[]' and '{}')

PR #884, opened 2018-07-31
fixed bug #152 (Hash Symbol in File Path prevents saving)

HomeofTrees:
This fixes bug #152 

changed all calls 
`StreamResult(File(String))` 
to 
`StreamResult(String)`

The solution might be the result of StreamResult corrupting Filepath by adding "file:%Filepath%" as pointed out here https://stackoverflow.com/questions/15324008/writing-xml-in-java-filenotfoundexception

tested with latest Version in Repo (v0.2.9.1)


Innoxia:
Thank you for this!

PR #889, opened 2018-08-05
Pull parsing out into its own namespace

Funtacles:

The intent here is to get parsing being used system wide in place of the string builders. This is the first step in that which is to pull parsing out of UtilText and put it into its own namespace and class for better maintainability.

   - Copy parser methods into a Parser class in its own namespace.
   - Move parser helper classes into that namespace
   - Simplify parser methods

*Does NOT* actually replace any of the existing parsing. That still lives in UtilText. This is just the first step in doing so.

Nyet

Tested on 0.2.9.5
Don't have one. Any communication can be handled through Github.


CognitiveMist:
@Funtacles 

This PR is out of date with most of the code it is based upon. Do you still believe it is useful? Do you think it might be worth updating?

Innoxia:

PR #904, opened 2018-08-18
Custom image accessibility

DJ4ddi:
So I originally intended to wait for some bug reports so I could just merge the fixes in with this, but for the sake of keeping it at least somewhat condensed I won't.

### Goals

This PR aims to further improve the accessibility and ease-of-use of custom images. This was already started with the ingame button to add images, now it is joined by its friend, the remove button.

Furthermore, images are now ex- and imported along with their characters. This applies to both the player (when starting a new game) and slaves (when importing at the auction). It should also make sharing characters with images easier, as people no longer need to look for them in the depths of the resource folder.

### Changes

The delete button shouldn't need much explaining. It removes the image, then reloads, that's all.

Import and export are also more or less self-explanatory. A directory with the same name as the .xml file containing the character information is created and all images of the character are copied into it. The copy is recursive, so artist folders are preserved.

Something that does need explaining is the new FileUtils class. I found that file handling was all over the place with very little error handling and the outdated Utils API. In order to streamline this, as well as to improve performance and error handling, I wrote some standard use cases (copy, delete) using the more modern NIO2 API and put them here. All methods are documented, if there are any questions I can be reached via Discord.

### Testing

All changes have been tested based on 0.2.10.2.


src/com/lilithsthrone/utils/FileUtils.java
```diff
@@ -0,0 +1,175 @@
...
+     * @param source The path of the source file
+     * @param destination The path of the destination file
+     * @param atomic Indicates if the operation should execute in a single step, which guarantees thread safety and
+     *               completeness, but is more likely to cause an error
+     * @return True if the operation succeeded, false otherwise
+     */
+    public static boolean copyFile(Path source, Path destination, boolean atomic) {
+        try {
+            // Skip existing directories
+            if (Files.isDirectory(destination) && Files.exists(destination)) {
```
picobyte: it looks like the .exists() check can go, or it should at least not be placed after the .isDirectory() check.

DJ4ddi: 1. I only want to skip if the target is a directory and exists. I do not want to skip existing files because they may be replaced. Therefore this statement does what it's supposed to.
2. When you enter your example path you would get an AccessDeniedException, but assuming you use a location that you can actually write to, relative paths work fine as those are properly resolved as part of the Path API.
3. Intentionally failing with more than one missing parent makes no sense to me. Could you explain why you think it should?
4. The try block is long because most of the Files.whatever calls throw the same exceptions. Since they all contribute to the same fail state, it makes sense not to split this up.

picobyte: I think to allow custom images is nice, but it's not right to allow writing files just anywhere.

1. If you say it is as intended, Files.isDirectory(destination) can only succeed if the directory exists, so the Files.exists(destination) test is redundant.
2. You should ensure the destination directory is a subdirectory of the game path. Even then, unless (nested) in a particular dedicated folder, it should fail if the file already exists. Where does this folder name come from? From an xml in an artwork package that you download from the net? So is this a potential security issue?
3. When is it sensible to allow writing files in new folders deeper nested than 1?
4. A late exception leaves nested directories behind, whereas an early exception does not. So it's not exactly a clean fail.

DJ4ddi: There seem to be a lot of misunderstandings here, so let me try to clear those up:

First of all, the FileUtils methods are designed for general purpose usage. The image import and export do use these new methods because they're convenient, but the point is to provide generally streamlined file utilities.

So what you are arguing for is essentially that no code in the entirety of LT should ever be allowed to write files outside of the installation directory. This does not seem like a useful limitation as it can be done by the system (user permissions) in a more reliable way. If you still think that it should be done, feel free to create your own [SecurityManager](https://docs.oracle.com/javase/8/docs/api/java/lang/SecurityManager.html) implementation, which the file methods will take into account by default.
And while we're on the subject of security: It is (in theory) possible to modify a save game to include a specifically formatted path as character name, which would then be used as part of the path created by the game, which could then be used to copy an infested JPG or PNG file that could cause a security issue when manually opening it with a program that does not have 10-year old exploits fixed. I'm not saying that it should be neglected, but by your argumentation virtually all programs installed on my PC pose a security threat because they are capable of copying files and opening images.

Next up, a dedicated copy method should definitely not always fail if the file already exists. Replacing it is desirable in many scenarios and can be prevented by checking the presence of the file beforehand.
There are numerous examples of directory trees deeper than two levels in LT. For example, an imported image would be put into /res/images/characters/generic/((charactername))/custom, out of which only the first part (/res) is guaranteed to exist and only the first three (/res/images/characters) exist by default. It should be mentioned that I didn't code it that way, I only followed the structure that the game already had.

And finally, a late fail during a file operation is never clean unless it is guaranteed to be atomic by the system (which is almost never the case). The process can fail at any time for a variety of reasons, leaving partially written files behind. You are correct that there may be empty directories under very specific circumstances, but those would require manual cleanup in any case, regardless of wether things are done in one try block or multiple. Thing is: I could see arguments in favor of always checking for incomplete files or empty directories and delete them, but for the purposes of debugging on both the developer and the user side (e.g. moving files manually) it may actually be desirable to have those folders created.
What I'm trying to say is this: Your point has nothing to do with a long try block and applies to any kind of writing file system operation.

On the positive side, the Files.exists call is indeed redundant, thanks for pointing that out.

picobyte: I sense a certain reluctance, but thank you for considering some security for a general purpose FileUtils methods.

Maybe with something like the isParent() test from here:
https://stackoverflow.com/questions/16186285/secure-way-to-create-relative-java-io-file
it is possible to limit the path to a subdirectory of the jar, obtained with something like this:
String f = new File(Main.class.getProtectionDomain().getCodeSource().getLocation().toURI()).getPath();
from here:
https://stackoverflow.com/questions/320542/how-to-get-the-path-of-a-running-jar-file

DJ4ddi:
### Update

After some more aggressive testing with various special characters in names (thanks to @picobyte for pointing it out) I managed to break both the internal directory structure and the parser multiple times. I therefore decided that it's time for a more widely adopted solution.

#### Changes

Commit 5a90a316b8fa1019e288a61ce6129d8eae51891b adds a new utility method to validate and fix names. This is now used to check enchantment, character and place names. When attempting to save an invalid name, an "Invalid name!" message will be flashed. When attempting to load an invalid name from an XML file, it will be quietly corrected.

This fixes a variety of bugs when naming things with characters such as [, ], #, ../, etc.

#### Testing

As usual, things have been tested on 0.2.10.2.

picobyte:
It is obvious that you spent a lot of work and addressed some potential pitfalls. Although I'd advocate more restrictions in validate() - besides stripping particular characters, impose a relative path to the game or a few associated folders - I'll give my ack, for what it's worth.

PR #917, opened 2018-09-03
Deduplicate code in the Pregnancy Stats screen. 

uglybead:

Deduplicate code in the Pregnancy Stats screen. Also adds a statistic for percentage of offspring met.

As in the purpose.

No.

Yes. Works at head.

n/a

### DIFF
--- a/src/com/lilithsthrone/game/dialogue/utils/PhoneDialogue.java
+++ b/src/com/lilithsthrone/game/dialogue/utils/PhoneDialogue.java
@@ -997,11 +997,61 @@ public class PhoneDialogue {
 	public static final DialogueNodeOld CHARACTER_STATS_PREGNANCY = new DialogueNodeOld("Pregnancy Stats", "", true) {
 		private static final long serialVersionUID = 1L;
 
+		private void OffspringHeaderDisplay(StringBuilder output, String word_one, String word_two, String color, int count) {
+			output.append("<div class='extraAttribute-quarter'>");
+			output.append(word_one);
+			output.append("<br/>");
+			output.append("<b style='color:").append(color).append(";'>");
+			  output.append(word_two);
+			output.append("</b>");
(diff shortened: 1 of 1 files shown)


src/com/lilithsthrone/game/dialogue/utils/PhoneDialogue.java
```diff
@@ -997,11 +997,61 @@ public DialogueNodeType getDialogueNodeType() {
 	public static final DialogueNodeOld CHARACTER_STATS_PREGNANCY = new DialogueNodeOld("Pregnancy Stats", "", true) {
 		private static final long serialVersionUID = 1L;
 
+		private void OffspringHeaderDisplay(StringBuilder output, String word_one, String word_two, String color, int count) {
+			output.append("<div class='extraAttribute-quarter'>");
+			output.append(word_one);
+			output.append("<br/>");
```
picobyte: shouldn't above line be `output.append("<br>");`? it was like this in the original code, but I think that may be wrong?

CognitiveMist: `br` is a self-closing tag (and a bit of a special case). Technically, `<br/>` is most correct, but historically `<br>` was used most often. Nearly all html renderers will treat both the same way. An actual closing tag would look like `</br>`, and would almost always be incorrect.

uglybead: In the existing codebase:

* the "<br>" form is used once (in com/lilithsthrone/game/dialogue/utils/OptionsDialogue.java)
* the "<br/>" form is used at least 2574 times (number of lines where it occurs)
* the "<br />" form is used zero times

It seems like what is done here aligns most with the existing codebase.

Innoxia:
Thank you for this!

PR #928, opened 2018-09-11
Add feminizing FluidEffect for cum/girlcum/milk

akio-lt:

        Add feminizing FluidEffect

        Added a FluidEffect called Feminizing that calls incrementFemininity() on imbiber.

        No

        I have tested that it works in basic cum / girlcum / milk scenarios.  v0.2.10.8 

        Akio on Discord

### DIFF
--- a/src/com/lilithsthrone/game/character/CharacterUtils.java
+++ b/src/com/lilithsthrone/game/character/CharacterUtils.java
@@ -1259,6 +1259,9 @@ public class CharacterUtils {
 					character.addMilkModifier(FluidModifier.FEMINIZING);
 				}
 				if(Math.random()<=0.025f) {
+					character.addMilkModifier(FluidModifier.MASCULINIZING);
+				}
+				if(Math.random()<=0.025f) {
 					character.addMilkModifier(FluidModifier.HALLUCINOGENIC);
 				}
 			}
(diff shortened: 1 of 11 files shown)


akio-lt:
I literally have no idea why I didn't think of that, ha. Agreed.

akio-lt:
Added Masculinizing effect as well. Also added better handling for when the imbiber wouldn't be affected by the femininity change.

src/com/lilithsthrone/game/character/CharacterUtils.java
```diff
@@ -1255,6 +1255,12 @@ public static void randomiseBody(GameCharacter character, boolean randomiseAge)
 				if(Math.random()<=0.025f) {
 					character.addMilkModifier(FluidModifier.ADDICTIVE);
 				}
+				if(Math.random()<=0.025f) {
+					character.addMilkModifier(FluidModifier.FEMINIZING);
+				}
+				if(Math.random()<=0.025f) {
+					character.addMilkModifier(FluidModifier.MASCULINIZING);
+				}
```
CognitiveMist: Should it be possible for an npc to generate with both modifiers? If so, perhaps both together would have a special effect, like a randomized change, or an androgynizing change (towards center), or an extremifying change (away from center).

akio-lt:
Hmmm.. valid point. Androgynizing for having both is probably the best solution. I was thinking about how it might be best to handle that.

src/com/lilithsthrone/game/character/CharacterUtils.java
akio-lt: Hmmm.. valid point. Androgynizing for having both is probably the best solution. I was thinking about how it might be best to handle that.

Innoxia:
This is very cool, and I really appreciate the time and effort you put into this, but I'm planning on adding enchantment effects to all bodily fluids, which will also include femininity modifications. As a result, I'm unwilling to merge this, as it will end up conflicting with the fluid effects once added. Sorry! ;_;

PR #930, opened 2018-09-12
Properly enslave offspring additionally to apply enslavement effects

orvail:
Tested manually.


Innoxia:
Thank you for the fix - you were quicker to this than I was. I've already got this sorted in 0.2.12, however, so I can close this now. ^^

PR #931, opened 2018-09-14
Localization

DJ4ddi:
### Goals

In its current state, the game is using mixed unit systems both internally and in the UI, leading to inconsistencies and unnecessary conversion.

In order to fix that, this PR localizes everything. Dates and numbers use the common local formats. Depending on the locale, the game picks between metric and imperial units, automatically converting volumes, weights and sizes.

One of the key design goals was to avoid any kind of internal unit calculation for stored variables. Every conversion is done when the HTML code for the UI is constructed. The only place that technically breaks this paradigm are status effects, as their attribute modifiers are cached long before they are used.

### Changes

The core is implemented in the new `Units` class. Various formatters are initialized on first use or manually using the `FORMATTER.updateFormats()` method. This must be after the settings have been loaded, which is why the initialization of attribute modifiers (which previously constructed the string immediately on object creation) is now delayed.

Three new settings determine the behavior of the formatters. When the automatic locale detection setting is enabled, all date, time and number formats as well as the imperial unit flag are derived from the system language. If set to manual, a 24-hour and an imperial unit toggle determine most of the formats and the language is defaulted to English.

These formats are then used in a variety of utility methods, all of which output strings for use in the HTML code:

- `time` only takes a `TemporalAccessor` and returns the formatted time of that point.
- `date`, in addition to the point in time, takes a `DateType` enum, which can be either `SHORT` for a compact, numeric format (e.g. for save game dates) or `LONG` for a spelled out format (e.g. for use in text).
- `dateTime` combines the two and outputs both date and time, concatenated with `, `
- `number` formats a single number. An overload with two integers to define minimum and maximum decimal places exists.

The actual unit methods `size`, `fluid` and `weight` each have 3 parameters (disregarding the convenience overloads that use the most common default): The value (`double`), the output type of the value (`ValueType`) and the type of the unit (`UnitType`).

The value type can be one of three parameters:
- `NUMERIC` for simple numbers, rounded to the nearest integer
- `PRECISE` for numbers with additional precision, which means 2 post decimal places for metric and eighths for imperial units. Precise values also are not wrapped in order to avoid precision loss.
- `TEXT` for full text numbers using the utility method `intToString` that is already present. Text numbers are rounded more aggressively in order to avoid getting too long, e.g. to the nearest 5 between 10 and 50 and the nearest 10 above 100.

The unit type has four options:
- `NONE` simply omits the unit and prevents wrapping, e.g. 500.
- `SHORT` prints an abbreviated format, e.g. 5 m.
- `LONG` uses the full text unit, e.g. 5 metres.
- `LONG_SINGULAR` also uses the full text unit, but it is concatenated with a dash and always singular, e.g. 5-metre.

All input values are given in the smallest available metric system unit, which means centimetres for sizes, millilitres for fluids and grams for weights. To that end, existing values have been converted internally to avoid having to build a two-way conversion system. **The numbers are converted on loading, the version must be updated after merging.** The place to change is marked with a `// FIXME`.

Finally, there are some new utility methods for rounding, e.g. printing a value with eighths, rounding to _n_ places, rounding to the nearest _n_ and adaptive rounding based on the input value. These are all safe to use outside of the actual unit calculations.

Most changes outside of the unit class are simple replacements for numbers so the appropriate unit function is used instead. For the most part, I stuck to the existing formats, so that the output is roughly the same.
In some places, adjustments to increment sizes were made, especially for fluids. This is because adding 1 mL doesn't only seem pointless, but it also causes a click (+1 mL) to lack a UI reaction due to rounding in the imperial system.

Some new parser tags allow variable units in text blocks. They take a value as an optional argument, which is converted to the appropriate unit. If no value is given, the name of the unit is printed instead. The new parser target `unit` is used to access them. Currently, there is:
- `size`, which prints "_inch_" or "_centimetre_" or a numeric, long, singular unit.
- `sizes` or `sizePlural`, which outputs "_inches_" or "_centimetres_" or a long, numeric unit.
- `lSize` or `largeSize` which outputs "_foot_" or "_metre_" or a long, singular text unit.
- `lSizes` or `largeSizePlural` which outputs "_feet_", "_metres_" or a long text unit.

Existing text occurences have also been replaced. Since there were no fluids or weights used in text, I didn't bother adding parser tags for those. Given the existing format, it should be easy to do that later, as required.

### Testing

All changes have been tested on 0.2.10.5. I did merge 0.2.10.8, but since it's a bit buggy I did not extensively test on that version.
Due to the amount of small changes made, there is a high chance that I missed something, somewhere.


src/com/lilithsthrone/game/Game.java
```diff
@@ -3573,11 +3494,6 @@ public void setResponsePointer(int responsePointer) {
 	public boolean isPlayerTileFull() {
 		return getActiveWorld().getCell(getPlayer().getLocation()).getInventory().isInventoryFull();
 	}
-
-
-	public boolean isDebugMode() {
-		return Main.getProperties().hasValue(PropertyValue.debugMode);
-	}
```
CognitiveMist: What's the purpose of this change?

DJ4ddi: It's the result of a bad merge. Thanks for the note, I'll clean it up.

ltpatch:
Personally, I think the change in units everywhere (including the save file) is too complicated/unnecessary. The date and number formatting is a nice, clean improvement, though.

DJ4ddi:
Would you mind elaborating on why you think it is too complicated when the work is already done?

PR #939, opened 2018-09-25
Character UIDs

DJ4ddi:
This PR is based on and therefore contains #904. I moved it to a seperate branch because it is less important and has more potential to break things.

### Goals

In their current iteration, character ids are only unique within one save game. This PR makes them locally unique instead (aka unique within this installation of Lilith's Throne).

This currently only has advantages for custom image handling, but could potentially be useful in the long term for other aspects that transcend save games (e.g. any kind of file handling).

### Changes

The new UID class handles generating locally unique identifiers using `UID.get()`. The resulting string consists of a 48 bit timestamp in milliseconds (slightly above 8925 years) and a 16 bit random integer, converted to a base 32 string for compactness. It can never be longer than 13 or shorter than 4 characters.

The timestamp is checked for duplicates, e.g. when two ids are generated within one millisecond. This check does not persist through restarts, which can potentially lead to duplicates in any of the following circumstances:
- The game is restarted and loaded within 1 ms.
- There are enough ids generated within 1 ms to trigger an overflow (around 280 trillion currently)
- The system clock is adjusted backwards while the game is not running.

Realistically, the last option is the only one worth considering. In that scenario, the same millisecond timestamp could come up twice. With 16 random bits, the chance of a duplicate is 0.0015%. This is certainly good enough for character ids, but should not be used for passwords or cryptographical applications.
\
\
The conversion process to the new ids happens upon loading a save game. In order to provide long term flexibility, the id map can now be used to store aliases (the same character with multiple ids). Duplicates are prevented at runtime by using a set instead of a list for `getAllNPCs()` and permanently by resolving ids with the map when saving the game.

This approach is more complicated than simply doing a batch conversion on load, but has the advantage that there are no invalid ids saved (which is something that currently happens a lot).

Character images are moved to the new folders immediately. To make sure that the new ids aren't lost, an autosave is triggered after loading.

### Testing

All changes have been tested with 0.2.10.8. It is possible to "lose" the character images by loading the same save twice (and therefore overriding the autosave). I consider that an acceptable issue as it would take a lot of extra work to prevent and alpha users are likely familiar with the folder structure (so they can clean it up).


orvail:
Please consider using a type 4 uuid instead of something hand-crafted.

DJ4ddi:
Considered and declined. All UUID variants are significantly longer than necessary, which is relevant since one of the main purposes is to use them in file names.

PR #940, opened 2018-09-27
SVG optimization

DJ4ddi:
Don't let the line count deceive you, this one isn't very complicated.

### Goals

Inkscape is a great tool to edit SV graphics. It also adds a ridiculous amount of clutter that is completely useless to the WebEngine renderer. This PR aims to reduce the size of SVGs, improving performance slightly across the board and significantly on icon-heavy displays (e.g. the inventory).

### Changes

There are only two relevant code changes:
1. Some regular expressions (most notably the gradient id replacement) were improved.
2. The short form of hex colors (#faa instead of #ffaaaa) can also be replaced.
3. The pattern layer is identified by group id (`<g id="patternLayer"`) instead of an Inkscape id.

The optimization was done in two steps: First, all `<flowRoot>` elements were removed manually. These are randomly added by Inkscape when dragging things in the editor and serve no purpose since they technically don't even exist in SVG 1.1, which all the files use.
Second, automated optimization was done using [SVGO](https://github.com/svg/svgo), a relatively simple command line tool. The configuration used is included ([here](https://github.com/Innoxia/liliths-throne-public/blob/83f67383dc826b097de428813dd93e82a00350f0/svgoConfig.yml)) and contains instructions on how to use it. The only files I used a different configuration for are the camo and multi-camo patterns, which for some reason had issues when moving attributes (`moveElemsAttrsToGroup`, `moveGroupAttrsToElems `and `removeUnknownsAndDefaults`).

There also exists an [Inkscape plugin](https://github.com/konsumer/inkscape-svgo) automating the process when exporting SVGs. I have not used it and can't give details about its configuration, but it may be worth looking into.

### Profiling

Here are some numbers measured on the same save game, identical VM, both warmed up and with clean file caches. Especially the last point needs to be taken into account when doing your own tests.

Note that this is only the time required for the WebEngine to parse the DOM tree and does not include rendering times, so the overall improvement is slightly bigger.

Vendor:

| Panel | Original | Optimized |
| ------------- |:-------------:|:-----:|
| Center | 2851 kb in 136 ms | 488 kb in 53 ms |
| Bottom | 53 kb in 3 ms | 12 kb in 2 ms |
| Left | 2185 kb in 89 ms | 288 kb in 26 ms |
| Right | 1014 kb in 47 ms | 177 kb in 16 ms |

Tile:

| Panel | Original | Optimized |
| ------------- |:-------------:|:-----:|
| Center | 14 kb in 14 ms | 6 kb in 14 ms |
| Bottom | 53 kb in 3 ms | 12 kb in 2 ms |
| Left | 756 kb in 35 ms | 162 kb in 19 ms |
| Right | 141 kb in 9 ms | 38 kb in 5 ms |

Overall, the size of all SVG files was reduced from 23.67 mb to 4.23 mb at no visual quality loss.

From those numbers and some other measurements not listed here we can see that the difference on regular tiles isn't huge and not very noticable at 20 ms on average. However, the difference in a vendor inventory is usually around 180 ms and often larger than 200 ms, which is a significant improvement (even without accounting for rendering times).

This could be further reduced by getting rid of SVGs entirely and using PNG images instead, which I started working on a while ago on [this branch](https://github.com/DJ4ddi/liliths-throne-public/tree/svg-conversion). This has the disadvantage of reducing the quality of scaling and doesn't allow color replacement. But since the changes there are a lot more complex, other performance issues are simply more relevant (most notably avoiding the complete reconstruction of the UI after every click).

### Testing

The changes have been tested on 0.2.10.8. I checked as many graphics as possible and found no issues, but it is possible that I missed some.


CognitiveMist:
I like this. Would it be feasible to discard some common extraneous elements when loading the svg? Specifically for use with modded items. We can't realistically expect every mod author to optimize their svgs, unless it is easy to do. (Maybe the Inkscape plugin would be easy-ish, but it still won't be used by many unless its enforced somehow.)

DJ4ddi:
Good question. Possible? Yes, definitely. Feasable? Hard to say.

As I see it, the process (given that the configuration is already provided) is simple enough for anyone to do, and even if an SVG gets submitted without it, one file won't make a huge difference and can be done by someone else as a batch (it takes seconds to process a whole folder).

Integrating the JavaScript library is doable and likely faster than reimplementing the features. The bigger issue I see is dealing with exceptions. With the manual way, making an exception by not processing the file or disabling some optimizations for it is trivial. In an automated process, there would need to be some sort of indicator for special cases.

In conclusion, it's probably better to point mod authors the right way and hope that at least some follow the guidelines. Enforcing it would be a significant effort and may or may not break things.

Innoxia:
Thank you! ^^

PR #949, opened 2018-10-23
Faster save loading

AlacoGit:
Thanks to Pimgd on the discord for fixing the race condition related NPEs

The main purpose of this patch is to speedup the loading of saves, especially ones with large amounts of characters.
The patch accomplishes this by using Javas `streams` API to parallelize loading of NPCs during save-loading. 
While the individual performance gains will wary between users, during testing an average gain of 50% was observed.

This is my first time really working with `streams` so any feedback is very welcome.

 Parts of `importGame` was rewritten to use the `streams` API, some other small misc changes was also made.

 A new method `public safeAddNPC(final NPC npc, boolean isImported)` was added, this class wraps the original `addNPC` to catch any exceptions thrown since lambdas can't contain exceptional methods.

 `npcTally` in `Game.java` was rewritten to use `AtomicInteger` instead of a regular int, this to prevent race conditions. Similarly the implementation of `NPCMap` was changed to `concurrentHashMap` for the same reasons.

 Large parts of the parsing logic in `UtilText` was thread-hostile and was refactored. As a part of this `ParserCommand.parse` was changed to also take a `GameCharacter` argument

 No

 Tested on 0.2.10.8 and the pre-release version of 0.2.11

 AlacoGit on Discord


src/com/lilithsthrone/game/Game.java
```diff
@@ -3402,15 +3409,15 @@ public String addNPC(NPC npc, boolean isImported) throws Exception {
 			}catch(NumberFormatException ex) {
 				tallyCount = Integer.parseInt(npc.getId().split(",")[0]);
 			}
-			if(tallyCount>npcTally) {
-				npcTally = tallyCount;
+			if(tallyCount>npcTally.get()) {
+				npcTally.set(tallyCount);
```
CognitiveMist: This is no longer atomic, since the get and set are separate. Perhaps something like this would work better:

```
npcTally.updateAndGet(x -> Math.max(x, tallyCount));
```

src/com/lilithsthrone/game/Game.java
```diff
@@ -986,18 +979,22 @@ private static NPC loadNPC(Document doc, Element e, String className,
...
-				Constructor<? extends NPC> declaredConstructor = constructorMap.get(npcClass);
-				NPC npc = declaredConstructor.newInstance(true);
-				loadFromXMLMethodMap.get(npcClass).invoke(npc, e, doc, new CharacterImportSetting[] {});
-				return npc;
 			}
+			Constructor<? extends NPC> declaredConstructor = constructorMap.get(npcClass);
+			NPC npc = declaredConstructor.newInstance(true);
+			loadFromXMLMethodMap.get(npcClass).invoke(npc, e, doc, new CharacterImportSetting[] {});
+			return npc;
+		} catch(ClassNotFoundException CNFE) {
```
CognitiveMist: All-caps names like `CNFE` or `NSME` are generally reserved for constants. Here, it would be more common to use `cnfe` and `nsme`, or similar.

src/com/lilithsthrone/game/dialogue/utils/UtilText.java
```diff
@@ -5233,10 +5233,10 @@ public String parse(String command, String arguments, String target) {
 		}
 	}
 
-	private static GameCharacter character;
```
CognitiveMist: I know this was necessary for what you did, but thanks for making this global var disappear. In my opinion LT has way too many globals.

src/com/lilithsthrone/game/Game.java
AlacoGit: You are correct, thanks for spotting that!

src/com/lilithsthrone/game/Game.java
```diff
@@ -1024,18 +1017,22 @@ private static NPC loadNPC(Document doc, Element e, String className,
...
-				m.invoke(npc, e, doc, new CharacterImportSetting[] {});
-				return npc;
-			} else {
-				Constructor<? extends NPC> declaredConstructor = constructorMap.get(npcClass);
-				NPC npc = declaredConstructor.newInstance(true);
-				loadFromXMLMethodMap.get(npcClass).invoke(npc, e, doc, new CharacterImportSetting[] {});
-				return npc;
 			}
+			Constructor<? extends NPC> declaredConstructor = constructorMap.get(npcClass);
+			NPC npc = declaredConstructor.newInstance(true);
```
Innoxia: This is throwing a NullPointerException, seemingly at random. I assume there's something going wrong with writing to the ConcurrentHashMaps, but I can't figure it out. :sweat_drops: 

Each time I try to load a game, it's a different class(es) that fails, and when it does, it's not for every NPC in that class, but just for one or two of them. Sometimes the loading works without any error being thrown at all. Sometimes it's as many as five (on my moderately-sized save file).

Innoxia:
Thank you! ^^

PR #960, opened 2018-11-05
Suggested refactoring method to reduce the number of switch statements

fetgal:

This pull request should result in no functional change. It is simply a code re-factor suggestion which could be used to reduce reliance on switch statements for enum functions.

Changed the Eye & EyeType class and added a class called EyeTransformationText. I pulled all the text from the switch statement in Eye into variables in the EyeTransformationText class to separate the game text from the code, making it easier to read the code.

Added properties to the EyeType enum, the most important being the transformationText property which is declared as Function<T, R>. This allows a function to be stored as a property to be called later. Each enum has to define the function, which allows the calling code to just invoke it, instead of using a switch to decide what to do.

No

I tested quickly on version 0.2.11

fetgal


src/com/lilithsthrone/game/character/body/types/EyeType.java
```diff
@@ -4,6 +4,9 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
+
+import org.h2.util.StringUtils;
```
AlacoGit: This package isn't part of the the standard JDK and thus we can't use it unless the dependency is added.

fetgal: Oops, that wasn't the one I wanted anyway.

I did notice the project has no dependencies. Is that a deliberate choice? I can certainly rework this without bringing in a dependency.

Besides my botched import, any thoughts?

AlacoGit:
I've had a chance to look at this more closely and it mostly looks good, however the way `getTransformationText()` works is very non-idiomatic java, almost C like. Is there a reason this can't be a regular function that gets overwritten by the individual members?

src/com/lilithsthrone/game/character/body/types/EyeType.java
```diff
@@ -95,65 +302,14 @@ public String getNamePlural(GameCharacter gc) {
...
-				return UtilText.returnStringAtRandom("fox-like");
-			case RAT_MORPH:
-				return UtilText.returnStringAtRandom("rat-like");
-			case RABBIT_MORPH:
-				return UtilText.returnStringAtRandom("rabbit-like");
-			case BAT_MORPH:
-				return UtilText.returnStringAtRandom("bat-like");
-		}
-		return "";
+		return UtilText.returnStringAtRandom(StringUtils.split(this.descriptor, ','));
```
CognitiveMist: You can use this to avoid the dependency:
```java
this.descriptor.split(",")
```
Or, if you feel that you need null-safety (which might still break returnStringAtRandom):
```java
this.descriptor != null ? this.descriptor.split(",") : null
```

src/com/lilithsthrone/game/character/body/types/EyeType.java
```diff
@@ -17,50 +20,254 @@
...
+	public EyeShape getIrisShape() {
+		return this.irisShape;
+	}
+	
+	public EyeShape getPupilShape() {
+		return this.pupilShape;
+	}
+	
+	public Function<Boolean, String> getTransformationText() {
+		return this.transformationText;
```
CognitiveMist: To build off AlacoGit's comment, I personally think this is okay Java, however, it might be rather unexpected. I would suggest moving the apply inside the getter unless there is a good reason to do it elsewise:
```java
public String getTransformationText(boolean isPc) {
    return this.transformationText.apply(isPc);
}
```

src/com/lilithsthrone/game/character/body/types/EyeType.java
fetgal: Right, in this instant I know the dependency doesn't buy me much. It is kind of habit for me to bring in this dependency, as most projects I've worked on have included it. Just kinda using it to figure out the dependency situation, since I was surprised to see none.

src/com/lilithsthrone/game/character/body/types/EyeType.java
fetgal: I liked your suggestion about moving the apply.

As far as Java goes, I think it feels unexpected because a lot of the functional programming elements in Java 8 are still uncommon to see. I've only seen this sort of thing used a couple times, and this is only my second implementation, but I think it makes the calling code easier to read and it isn't to hard to figure out what it is doing if you have to look at the methods.

fetgal:
Unrelated to everything else mentioned I think there might be a bug in the code here, but didn't want to make the final call. There appears to be no text for a player character to receive DEMON_COMMON eyes. The boolean call is used to differentiate between imps and demons, not pc and npc.

Rfpnj:
I have a tendency to look at the number of lines of code in any refactor, and in this case what I am seeing is an increase of 81 lines of code over what it has currently. Seeing how the description talks of removing switch statements you would think you would see an overall reduction in the number of lines. Well, it is nice to look at reducing the number of switch statements in the code, in this case, I think in the long run it would add more work when adding new races.

fetgal:
@Rfpnj Looking at the number of lines in a code refactor is fine, but there are a lot of way to mess with that number. I like to see each of a constructors parameters on a new line after it reaches a certain length. That adds a lot of lines, not a lot of complexity. I trimmed down my new lines on my constructors and now I'm -21 lines for the refactor.

As for adding more work when adding new races, that is incorrect. Currently if I wanted to add a race, I'd have a modify the Eye file and the EyeType file. I believe right now I'll get help from the compiler since there is no default in the switch statement, but if one snuck in I'd loose that help and end up with a bug if I forget to add it to the switch.

With the refactor adding a race is as simple as adding a enum. 1 addition to 1 file, the definition of a simple change.

src/com/lilithsthrone/game/character/body/types/EyeType.java
CognitiveMist: I tend to look at this from the opposite angle, in that I avoid dependencies unless it provides a significant benefit, whether it be efficiency (number crunching) or correctness (crypto) or simplicity (lombok). To me, more dependencies means more to maintain, to update versions or make an artifactory or whatnot. I may be getting ahead of things a bit, but I don't think this really warrants the bother.

However, this isn't my call to make, it's Innoxia's. If she's cool with it, so am I.

fetgal: I certainly won't disagree with you that too many dependencies are a bad thing and that they should be vetted. I've spent many an hour in dependency hell and know it is no fun. Happy to make the change if that is what is requested.

GeneralPikaru:
Regardless, it might be better for text to be stored elsewhere. its quite distracting to have Butts on the screen (literally literally) when trying to figure something out. From experience, and reagrding the nature of this Project, Id suggest writing the Content into files (XML seems to be preffered here) and when putting out Releases to minimize them into Bytecode. Runtime, they can then on Startup be reconverted and loaded into RAM and used, later to be discarded. As in the example with EyeTransformation, this reduces redundancies significantly. I can only approve of this.

PR #967, opened 2018-11-21
Fix instances of coreAtributes so that people copying existing xmls will not copy mistakes

Pimvgd:
Should have been done ages ago but whatever


Innoxia:
Thank you!

PR #945, opened 2018-10-04
More detailed error messages during mod loading

CognitiveMist:
There are a number of errors that can occur when importing a mod file
which give little idea as to what actually went wrong. This change
adds the filename to importing AbstractClothingType, AbstractTattooType,
and AbstractWeaponType, and adds the exception message to each error
case. Hopefully this will help mod authors figure out what exactly the
problem is.

### Things to include in a large Pull Request:

- What is the purpose of the pull request?
More detailed error reporting when loading mod files

- Give a brief description of what you changed or added.
Added detail to error messages; filename and exception message

- Are any new graphical assets required?
No

- Has this change been tested? If so, mention the version number that the test was based on.
Yes, v0.2.10.8

- So we have a better idea of who you are, what is your Discord Handle?
CognitiveMist|Phlarx

- If you want quick feedback, you can use @Innoxia, or jump on the Lilith's Throne discord and send Innoxia a PM.

PR #968, opened 2018-11-21
Improve error messages for loading Blocked Parts

Pimvgd:
Improving the error messages should lead to less debugging headaches for mod creators and lead to more-up-to-date mods.


Pimvgd:
Humm... related to https://github.com/Innoxia/liliths-throne-public/pull/945.

PR #969, opened 2018-11-21
Improve mod error descriptions

Pimvgd:
Based on #945 and #968. Feel free to read their descriptions. If this is merged, those can both be closed.


Pimvgd:
I think it's fine now, but it's totally different.

Innoxia:
Thank you for this! ^^

PR #981, opened 2018-12-04
prelim_path_fix

AlacoGit:
This is a potential fix to the classpath problem some users are experiencing.
Need people who build with Eclipse to try and see if it works for them.


Questiondeca:
Can confirm that this preliminary fix does correct the issue causing code 1f, as a note a blank error.log file is still created in data upon opening the Lilith's Throne jar.

Innoxia:
Thank you!

PR #988, opened 2018-12-11
Fix for imageEquippedName

CognitiveMist:
The path portion of imageEquippedName was left out. Also, don't
have two separate image entries if imageName and imageEquippedName
are the same.

As mentioned by Kobolds here: https://discordapp.com/channels/302617912949211136/446038411791433729/521647452139290624

Tested on version: 0.2.12.96
Discord handle: CognitiveMist|Phlarx


Innoxia:
Thank you for this! I'd fixed it locally in my code, but your fix is far better and more concise.

PR #989, opened 2018-12-11
Increased variety for Name.java

GynoidDoesGames:
Increased the name lists by a considerable degree, increased the variety (fewer Carls!), and made the "prostitute" name generator occasionally produce "regular human" names (50% of the time) and vice versa (10% of the time). That last change can be undone if it's an issue, like if it'd be better to handle that elsewhere in the code.

No new graphical assets required.

(Gynoid / GearheadLydia#8498 on Discord)

### DIFF
--- a/src/com/lilithsthrone/game/character/persona/Name.java
+++ b/src/com/lilithsthrone/game/character/persona/Name.java
@@ -15,71 +15,126 @@ import com.lilithsthrone.utils.Util;
  * @author Innoxia
  */
 public class Name {
-
+	// Some help from behindthename.com's name lists to find unusual forms and/or same-letter names.
+	// Name etymologies / explanations omitted to comply with their terms of use about redistributing their article contents.
 	private static List<NameTriplet> human = (Util.newArrayListOfValues(
 					new NameTriplet("Alexander", "Alex", "Alexandria"),
 					new NameTriplet("Alexius", "Alex", "Alexia"),
 					new NameTriplet("Alex", "Alex", "Alex"),
 					new NameTriplet("Ash", "Ashe", "Ashley"),
 					
-					new NameTriplet("Bart", "Bailey", "Bridget"),
-					new NameTriplet("Ben", "Beverly", "Bella"),
+					new NameTriplet("Bart", "Bailey", "Barbara"),
+					new NameTriplet("Ben", "Bennie", "Bella"),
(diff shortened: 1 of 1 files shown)


src/com/lilithsthrone/game/character/persona/Name.java
```diff
@@ -665,6 +732,11 @@ public static String getSurname(GameCharacter gc) {
 	
 	public static NameTriplet getRandomTriplet(Race r) {
 		NameTriplet name = Util.randomItemFrom(human);
+		// occasionally throw some "prostitute" names in there - about 10% of the time
+		if(Util.random.nextInt(9)>8)
```
Innoxia: The `nextInt(int bound)` method treats the specified value as exclusive, meaning that in this instance, the method will always return false. (As the returned values are 0-8, inclusive.)

I personally prefer to use Math.random() for percentage-based random chances, as I find it clearer to read. As such, I'd change this line from:
`if(Util.random.nextInt(9)>8)`
to
`if(Math.random()<0.1)`

src/com/lilithsthrone/game/character/persona/Name.java
```diff
@@ -724,6 +796,10 @@ private static NameTriplet getDemonName() {
 	}
 	
 	public static NameTriplet getRandomProstituteTriplet() {
-		return Util.randomItemFrom(prostitute);
+		// occasionally throw some "regular" names in there - 25% of the time
+		if(Util.random.nextInt(3)<3)
```
Innoxia: This will always return true, as `Util.random.nextInt(3)` returns a number in the range of 0-2, inclusive.

As in the other comment I made higher up, I would find it clearer if it were changed to:
`if(Math.random()<0.75)`

Innoxia: Also, I prefer all `if` `else` code blocks to be wrapped in parentheses, as the code can end up looking a little messy without them. ^^

GynoidDoesGames:
Also, it may be worth considering some changes that allow these name lists to be pulled from an external file, to make it easier to modify the lists (i.e., without recompiling the game each time). However, that's a much more involved set of changes, so I may have to take a swing at that later.

Innoxia:
I accidentally credited the wrong person in the patch notes for this, sorry, GynoidDoesGames! I've corrected it in the blog/patreon patch notes posts, and will add a line in the next release's patch notes about the correction.

PR #992, opened 2018-12-16
Added egglayer property

Ettanin1987:
* Purpose: Implement proper detection for egg-laying race.
Added an egg-layer property to the races definition and adjusted the egg birth scene at Lilaya's home to consider egg-layer property than just harpy only.
* No graphical assets required
* Tested on 0.2.12.96 (master) using a human, a harpy and an alligator.
* Discord handle: Ettanin#5020

### DIFF
--- a/src/com/lilithsthrone/game/character/race/Race.java
+++ b/src/com/lilithsthrone/game/character/race/Race.java
@@ -30,6 +30,7 @@ public enum Race {
 			Attribute.RESISTANCE_HUMAN,
 			FurryPreference.NORMAL,
 			FurryPreference.NORMAL,
+			false,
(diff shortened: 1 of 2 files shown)


GeneralPikaru:
This looks good, although it may be conflicting with, say, insect-races like bees, for example. For now however, this Looks great. good job!

Ettanin1987:
@DarkerDaysTeam Feel free to copy and improve this snippet. The problem is, I understand where the changes are needed, but I didn't have the time to test and maintain it.

Furthermore, it targets the wrong branch (In hindsight, sent the PR to release instead of dev)
I'll thus close the PR.

PR #994, opened 2018-12-17
Temporary fix for softlock on doms performing a second action in nightlife

Darkon47:


Darkon47:
My mistake, this does not actually fix the error, simply delays it

PR #995, opened 2018-12-17
nightclubNPCs now always use their own location

AlacoGit:

Fix two bugs involving the generic nightclub NPCs:
  1.Dom NPCs causing "IndexOutOfBounds" exceptions
  2.Sub NPCs not leaving their current cell when moving to a new cell

In the current version(0.2.12.96) dom NPCs will crash the game and sub NPCs don't remove their ID from their current cell when moving to a new one, this is technically already fixed in the current version however that fix is also the cause for bug # 1 above. This PR fixes both bugs. 
No
tested on 0.2.12.96 master branch 
Alaco on Discord


AlacoGit:
This issue seems to have been fixed in 3.0. Closing.

PR #998, opened 2018-12-26
Fix for perk point weirdness

BDPMP:
No longer given additional perk points on level up, instead relying on the getPerkPointsAtLevel function added in 0.3. This means that gameCharacters should no longer receive double perk points on level up.

Now saving perk point value using getAdditionalPerkPoints instead of getPerkPoints, as this saves the value that the load function is expecting. This means that the perk point value will no longer change on save/load (this was causing negative perk points, but it could also cause absurdly inflated perk points)

Tested with a fresh character on version 0.3:
Perk points were at the value expected for each level, and we unaffected by saving or loading.

### DIFF
--- a/src/com/lilithsthrone/game/character/GameCharacter.java
+++ b/src/com/lilithsthrone/game/character/GameCharacter.java
@@ -685,7 +685,7 @@ public abstract class GameCharacter implements XMLSaving {
 		
 
 		CharacterUtils.createXMLElementWithValue(doc, characterCoreInfo, "experience", String.valueOf(this.getExperience()));
-		CharacterUtils.createXMLElementWithValue(doc, characterCoreInfo, "perkPoints", String.valueOf(this.getPerkPoints()));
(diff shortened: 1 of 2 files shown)


GeneralPikaru:
In Theory, this works, but did you test it?
Anyway, this method requires a lot more method calls than necessary.
Please consider this when making fixes, because wonky fixes will require more fixes in the future.

BDPMP:
The main change (perk points based on level) was introduced in 0.3, I just removed the old code that was still giving additional perk points as part of leveling up, and updated the save function to save the number of additional perk points that a gameCharacter has, which is what the load function is expecting.

Admittedly, the description I gave is quite ambiguous, but if you check the changes, you'll see I've altered very little.

Innoxia:
Thank you for this! I've merged this locally, so it will show up as merged here on github when I push 0.3..0.5 later tonight. ^^

PR #1003, opened 2018-12-27
Update GameCharacter.java

GeneralPikaru:
Bugfix for negative perk points and Sadists applying lust damage to enemy instead to self, including
General rework of sadist/Masochist application in combat

the perk points one is supposed to have is now stored in GameCharacter.perkPoints, this includes
perk points added via debug menue. perkPoints increase only at levelup and via debug menue.
since XML loading restored the value directly into perkPoints, only XML saving was slightly altered,
to simply commit perkPoints to the file instead of calling a method to calculate.
(this may induce problems with char ports from anything 0.3 or older)
GameCharacter.incrementHealth() was entirely re-written. I hope you didn't write this yourself,
because I have to say, it wasn't aynthing but cleaning a mess someone left behind.
I reused most of the Code but rearranged it so that
a) it does what it is supposed to
b) is decently readable
c) it works with a lot less effort. (less computing required => smoother gameplay)
it was tested in 0.3, and found to be working.

My handle is Pikaru#9100

### DIFF
--- a/src/com/lilithsthrone/game/character/GameCharacter.java
+++ b/src/com/lilithsthrone/game/character/GameCharacter.java
@@ -685,7 +685,7 @@ public abstract class GameCharacter implements XMLSaving {
 		
 
 		CharacterUtils.createXMLElementWithValue(doc, characterCoreInfo, "experience", String.valueOf(this.getExperience()));
-		CharacterUtils.createXMLElementWithValue(doc, characterCoreInfo, "perkPoints", String.valueOf(this.getPerkPoints()));
+		CharacterUtils.createXMLElementWithValue(doc, characterCoreInfo, "perkPoints", String.valueOf(perkPoints));
 
 		CharacterUtils.createXMLElementWithValue(doc, characterCoreInfo, "health", String.valueOf(this.getHealth()));
 		CharacterUtils.createXMLElementWithValue(doc, characterCoreInfo, "mana", String.valueOf(this.getMana()));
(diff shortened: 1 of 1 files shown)


src/com/lilithsthrone/game/character/GameCharacter.java
```diff
@@ -11645,61 +11640,49 @@ public String incrementHealth(float increment) {
...
-				manaLoss = Attack.getModifiedDamage(null, attacker, Attack.SEDUCTION, DamageType.LUST, manaLoss);
-				
-				attacker.incrementFetishExperience(Fetish.FETISH_SADIST, 2);
-
-				this.setHealth(getHealth() + (increment*1.05f));
-
-				if(increment<0) {
-					Combat.incrementTotalDamageTaken(this, -increment*1.05f);
+				//Attacker is Sadistic
+				if (attacker.hasFetish(Fetish.FETISH_SADIST) && attacker != null) {
```
AlacoGit: This should be the other way around, first check if the attacker exists and then if it has the fetish.

GeneralPikaru: this does not matter to neither runtime nor Compiler.
both are checked separately and then logically AND added.

AlacoGit: It does matter because the `&&` operator is left-associative and short-circuiting.
This means that the left-most condition is evaluated first and subsequent conditions are only evaluated if the previous one was `true`.
In this case it means that the code checks if `attacker.hasFetish(Fetish.FETISH_SADIST)` is true AND THEN checks if `attacker != null` or in other words, first it checks if the attacker has the fetish and then if it exists.

You can easily make a test case for this if you need further evidence.

0d721077:
After merging this, elementals don't have any perk points anymore. Not even after leveling up. 
That was not intentional, was it?

GeneralPikaru:
@0d721077 
As far as I know, companions / elementals not having perk points should not be related to this change.
If it is related, however, then i highly suggest altering the code within NPC files.
In case you havent read the change, it effectively alters the function to return the amount of Perkpoints any Character (including NPCs) has, regarding both spent Perkpoints and initial Perks. NPCs should invoke the same method, getting the same results. I can look into this issue.

PR #1006, opened 2018-12-28
Fixed re-rolling of gender for half demons

ChillaChris:
Enabled the method generateHalfDemonBody to accept the base body's starting gender as a parameter.


Innoxia:
Thank you for taking the time to do this, but I'd already fixed it on my end before seeing this PR. :(

Thank you for your time and effort, and sorry that I didn't end up merging it!

PR #1024, opened 2019-01-06
Alter Tooltip to respect renamed coverable areas

0d721077:
Changelog for 0.3:
> Changed transformation references of 'breasts' and 'ass' to 'nipples'
> and 'anus', as that's all the transformation is affecting. (Breast
> and ass coverings are based on your torso's covering.)

Since 0.3 the body part types for ass and breast should actually
describe the appearance of the anus and nipples. The npc tooltip
about the body part types still revealed nipples and anus type even if
those areas are covered and unknown by the player.


Innoxia:
Thank you for this! It should show up as merged when I push 0.3.0.5 later today. ^^

PR #1032, opened 2019-02-02
Animated GIF support

DJ4ddi:
### Goals

This is a relatively simple PR to allow adding animated GIFs as character images.

### Changes

It is now possible to select .gif files when adding character images. Mostly due to performance considerations, they are loaded slightly differently than static images:
- Only the first frame is used for tooltips (prevents lag when hovering)
- Only the first frame is used if the file is larger than 10 MB (prevents huge files filling cache and memory); A warning is printed to the console
- The animation is not rescaled before caching (prevents long loading times as we'd have to rescale each frame separately and then reassemble the animation)
- Warnings have been added to the image readme and button description

All of these were implemented due to the simple fact that animations can have completely unpredictable sizes depending on their length. Unlike static images, which have a very predictable size after scaling, resizing animations is not as simple and due to the lack of support from native libraries not worth implementing.
To avoid memory issues, overloading the cache and significant lag when opening character pages with animations, the above restrictions are always applied.

### Testing

All changes have been tested with 0.3.0.

### DIFF
--- a/res/images/characters/modding.txt
+++ b/res/images/characters/modding.txt
@@ -1,6 +1,7 @@
 ----- CUSTOM ARTWORK GUIDE -----
 
 All images must be in PNG (.png) or JPEG (.jpg) format and are scaled down internally to at most 600x600.
+Animations in GIF (.gif) format are also supported. Note that large GIFs may slow down the game, which is why the animation is not played in tooltips and may not be over 10 MB in size.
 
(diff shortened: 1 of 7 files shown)


Innoxia:
Thank you for this! I've merged it locally, so it will show up as merged on here as soon as I push my code to github (in an hour or so). ^^

PR #1039, opened 2019-02-04
Fixes a Lilaya Condom Breaking reaction then/else error

EPSIL0N:
Swaps a then and else that were reversed, Lilaya was reacting to the condom breaking when it was a scenario where you simply didn't wear one.


Innoxia:
Thank you! ^^

PR #1051, opened 2019-02-11
Fixes for crotch-boob weirdness

GynoidDoesGames:
Some quick fixes to crotch-boobs to reduce some of the more obvious weirdness. In particular, stops crotch-boobs from being added to characters upon load, and allows crotch-boobs to show up in the description as long as they are enabled whatsoever in the options.

@Innoxia


src/com/lilithsthrone/game/character/GameCharacter.java
```diff
@@ -18081,8 +18081,8 @@ public String removeMilkModifier(FluidModifier fluidModifier) {
 	// Misc:
 	public boolean hasBreastsCrotch() {
 		if(Main.getProperties().udders==0
-				|| (this.getLegConfiguration()==LegConfiguration.BIPEDAL && Main.getProperties().udders==1)
-				|| (this.getLegConfiguration()==LegConfiguration.BIPEDAL && this.getRaceStage()!=RaceStage.GREATER)) {
```
Innoxia: These lines were here to make sure that players with crotch-boobs turned off in the options would never see their descriptions. Surely by commenting this out, it disables that behaviour, thus making it so all players see crotch-boobs regardless of settings?

src/com/lilithsthrone/game/character/body/Body.java
```diff
@@ -1499,10 +1499,13 @@ public static Body loadFromXML(StringBuilder log, Element parentElement, Documen
 			}
 			
 			AbstractBreastType crotchBoobType = BreastType.getBreastTypeFromId(breasts.getAttribute("type"));
-			if(Main.isVersionOlderThan(Game.loadingVersion, "0.3.0.6")
+			
+			// disabled because it was erroneously applying crotch boobs to every non-human on load
+			/*if(Main.isVersionOlderThan(Game.loadingVersion, "0.3.0.6")
```
Innoxia: I think I've fixed this on my end already, by changing it to:
`if(Main.isVersionOlderThan(Game.loadingVersion, "0.3.0.6") && importedLeg.getLegConfiguration().isBipedalPositionedCrotchBoobs()) { // Reset crotch-boob type as I accidentally applied crotch-boobs to demons
	if(body.isFeminine()) {
		crotchBoobType = RacialBody.valueOfRace(body.getRace()).getBreastCrotchType();
	} else {
		crotchBoobType = BreastType.NONE;
	}
}`

src/com/lilithsthrone/game/character/GameCharacter.java
GynoidDoesGames: @Innoxia True, but something is nonetheless wrong with it, because - as written - there is no setting that allows crotch-boobs to be visible on any character who is not a Greater Furry. This means that crotch-boobs spuriously disappear off of both Lesser Furries and Hybrids.

Innoxia:
I've changed the handling of crotch boobs for 0.3.2, which should solve this issue. Thank you for the feedback!

PR #1057, opened 2019-02-12
reference errors

EPSIL0N:
orifice reference typo


EPSIL0N:
First commit was that pussy creampie was referencing asshole

Innoxia:
Thank you very much for this!

PR #1064, opened 2019-02-18
Update pattern colors to use the pattern-specific colors

CognitiveMist:
Update pattern colors to use the pattern-specific colors instead of the clothing colors.

This allows the patternLayer to replace a secondary-color region in a clothing item, without the primary coloration now controlling two regions. This also allows the clothing colors to be adjusted without affecting the pattern colors.

The result of these changes are that the background of the pattern is now controlled by the secondary or tertiary pattern color, if unused. The multi-camo pattern is the only exception to that rule.

No, adjusted existing assets only

Yes, 0.3.0.7

CognitiveMist


Innoxia:
Thank you!

Issue #1067, opened 2019-02-23
0.3.1 (dev branch) has ConcurrentModificationException when doing initial Properties.savePropertiesAsXML()


arpie-steele:
Found on Mac with both HotSpot and the Oracle 1.8.0_201.

The issue appears to be updating Main.getProperties().setValue(....) in the JavaFX thread when another thread is doing Main.getProperties().savePropertiesAsXML(...) which uses an iterator on an unsynchronized. Then you get an error log like:
```
Version: 0.3.1
Java: 1.8.0_201
java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.sun.javafx.application.LauncherImpl.launchApplicationWithArgs(LauncherImpl.java:389)
	at com.sun.javafx.application.LauncherImpl.launchApplication(LauncherImpl.java:328)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at sun.launcher.LauncherHelper$FXHelper.main(LauncherHelper.java:767)
Caused by: java.util.ConcurrentModificationException
	at java.util.HashMap$HashIterator.nextNode(HashMap.java:1445)
	at java.util.HashMap$KeyIterator.next(HashMap.java:1469)
	at com.lilithsthrone.game.Properties.savePropertiesAsXML(Properties.java:222)
	at com.lilithsthrone.main.Main.main(Main.java:806)
	... 11 more
```

A terrible fix would be to fix this particular condition by gluing the utils.Units and game.Properties classes together.
```diff
@@ -17,6 +17,7 @@ import com.lilithsthrone.game.inventory.weapon.AbstractWeaponType;
 import com.lilithsthrone.game.inventory.weapon.WeaponType;
 import com.lilithsthrone.game.settings.*;
[...]

PR #1071, opened 2019-02-24
Issue 1067 Fix ConcurrentModificationException while iterating over Properties.values

arpie-steele:
This fixes game-ending ConcurrentModificationException seen on 2 different JVMs on a MacBook Air at initial writing of data/properties.xml when Properties.setValue() was called by one thread while Properties.savePropertiesAsXML() was still iterating over Properties.values.

The key change is we loop over an enum possible values instead of the mutable HashSet.

I didn't see this issue #1067  in 0.3.0.7 official builds, but it broke my attempts to test my own builds of 0.3.1

No new graphical assets.
Tested in 0.3.1


Innoxia:
Thank you once again! ^^

PR #1077, opened 2019-02-28
Fix for over-eager comment removal and other rendering foibles

CognitiveMist:
The game's text parser recently got the ability to parse out comments,
denoted by //, but it was a bit overzealous. It would consume things that
shouldn't be comments, such as urls like http://lilithsthrone.blogspot.co.uk,
leaving just http: hanging, and obliterating everything else on that line.
This caused the start screen to look weird, and nearly every svg to be rendered
at the wrong size.

I've put in a fix for this by adding a negative lookbehind for ':'. Other
options could be to require comments are on their own line (not ideal),
comments begin with ' //' (that is a space, then two slashes), or similar,
but I think this is the least restrictive option.

This issue was introduced in commit f04832cb. Bad Inno, you should be
doing smaller commits more often, even if you don't push them to Github.

I have tested this fix in my own environment, which is 0.3.1.01. My discord
handle is CognitiveMist|Phlarx.

### DIFF
--- a/src/com/lilithsthrone/game/dialogue/utils/UtilText.java
+++ b/src/com/lilithsthrone/game/dialogue/utils/UtilText.java
@@ -677,7 +677,7 @@ public class UtilText {
 				parserVariableCalls.add(matcherVAR.group().replaceAll("#VAR", "").replaceAll("#ENDVAR", ""));
 			}
 			input = input.replaceAll("(?s)#VAR(.*?)#ENDVAR", "");
-			input = input.replaceAll("\\/\\/(.*?)\n", "\n"); // Replace comments
+			input = input.replaceAll("(?<!:)//(.*?)\n", "\n"); // Replace comments (but not URLs, like http://)
 		}
 		
 		try {
--- a/src/com/lilithsthrone/game/dialogue/utils/UtilText.java
+++ b/src/com/lilithsthrone/game/dialogue/utils/UtilText.java
@@ -406,7 +406,7 @@ public class UtilText {
 
 	public static String getCurrencySymbol() {
 //		return "&#9679;"; // Circle
-		return "&#164"; // 'Generic' currency symbol
+		return "&#164;"; // 'Generic' currency symbol
 	}
 	
 	public static String getPentagramSymbol() {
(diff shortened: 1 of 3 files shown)


CognitiveMist:
Turns out there was more to fix than just the comment parsing; I still had issues on the inventory screen after the previous commit. After some tracking-down, I've found and corrected a few issues:

- The currency symbol `&#164;` was missing its semicolon
- The background style of the inventory panel had escaped its quotes; I put is back and changed it from blue to dark grey (the same color as behind the individual item cells).
- Some svg images, especially custom ones, had xml definitions like `<?xml version="1.0" encoding="UTF-8" standalone="no"?>` at the top. This was causing big issues for me all of a sudden. So, they're filtered out now. (And really, they don't need to be in the svg files in the first place, but it seems to be a part of an inkscape header?)

Image references; before, after just the first commit, and after both commits:

before
during
after

Innoxia:
Thank you so much for this! For whatever reason, I wasn't seeing the inventory rendering bugs, so it would have been difficult for me to track this one down. This really is a huge help - thank you again!

P.S. I will try to make smaller, more frequent commits in the future.

bicobus:
> Thank you so much for this! For whatever reason, I wasn't seeing the inventory rendering bugs, so it would have been difficult for me to track this one down. This really is a huge help - thank you again!

The jar file you distributed didn't have any of those problems, which is quite astouding. Would be nice to know why.

Innoxia:
I think that it might have something to do with the java version I used to create it (1.8.0_172). I seem to recall that when I tried a newer version of Java a few months ago, it detected that some of my inline css, which was working on 172, was incorrect (now fixed), and therefore did not render it as I intended. 172, while technically not handling these errors correctly, was resulting in the final rendered product behaving as I expected.

This issue of UI elements not working correctly, but working on 172, might have been due to the same issue?

bicobus:
The jar file I obtain by compiling the master branch is drastically different than the one you distributed. In any case, it is concerning that a regex (and logic) would behave differently between 172 and more up to date JREs. You should expect to face similar problems moving forward, therefore compiling and testing against the latest JRE might be advisable.

I also don't understand why the xml headers are suddently creating problems, as the svg files didn't change between versions nor did the version of my JRE.

CognitiveMist:
That also confuses me; I looked around for any changes to JavaFX config in the repo, etc., but came up dry. I don't really have a good explanation for this sudden change.

Innoxia:
The only thing I can think of that might have caused this is that I installed the most recent Java 8 version on my machine a few days ago (jdk1.8.0_202). This made it the default Java version in Eclipse, but I changed it back to 172 for the LilithsThrone project (due to 202 being bugged and unable to render svg gradients), so am unsure of how it would have affected it. :s

PR #1082, opened 2019-02-28
Unit polishing

DJ4ddi:
This is a follow up for the recently merged metric units.

### Changes

- Eights have been reduced to quarters because fractionals are hard
- Long imperial sizes are now wrapped (e.g. "7 feet and 2 inches" instead of "86 inches")
- Units can now be configured seperately (because partial metrication is still a thing, shoutout to the UK) within an own setting dialogue
- Fixed the missing fluid conversion on the character info page
- Fixed the large limit increase button not working

### Testing

All changes have been tested on 0.3.1.01.


Innoxia:
Thank you for this! ^^

PR #1101, opened 2019-03-11
Configurable chance for randomized body pref / forced-tf race

Eliria:
_This allows future in-client settings or pref file editing to adjust the randomization of npc forced-tf potions which I find tends toward same-race too often. Sometimes I want to spawn humans that prefer furry races in their tfs or want more surprise. Doesn't actually alter the probability. I would add an option to set the variable but not yet gotten around to figuring out the options menus._

Adds in configurable probability of the the racial body pref for NPCs being randomized (and therefor race of forced-tf potions). Saves/loads a float in the properties file as the percentage chance of a randomized race. Default still uses 15% random, 85% same-race.

New Properties config variable ( randomRacePercentage default 0.15f ), saving and loading in properties.xml, getter for variable ( getRandomRacePercentage() ), and using variable for randomized race body pref generation.

Tested and appears to work without issue.

- So we have a better idea of who you are...
Eliria

### DIFF
--- a/src/com/lilithsthrone/game/Properties.java
+++ b/src/com/lilithsthrone/game/Properties.java
@@ -73,6 +73,8 @@ public class Properties {
 	public int forcedTFPercentage = 40;
 	public int forcedFetishPercentage = 0;
 
+	public float randomRacePercentage = 0.15f;
+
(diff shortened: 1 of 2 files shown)


Innoxia:
Thank you!

PR #1102, opened 2019-03-13
Unit fixes

DJ4ddi:
And another one. Honestly, at 5000 lines changed there were bound to be some inconveniences. But since it's all about convenience, those shall also be fixed.

### Changes
- Fix missing sign of negative values
- Fix rounding issues (most notably for imperial sizes)
- Use plural for value 0
- Wrap long singular inches to \*-foot-\* (recommended to me as more readable than something like 80-inch, I have never used the imperial system so I wouldn't know)

### Testing
All changes have been tested on 0.3.1.1.


Innoxia:
Thank you once again!

Issue #1104, opened 2019-03-15
NullPointerException when approaching imp fortress if player's subspeciesOverride is null


Niranufoti:
The problem appears to be caused by [a recent change replacing `==` equality checks with `.equals()`](https://github.com/Innoxia/liliths-throne-public/commit/1248211e915ecb2e87e7b8fc60fddeeaa25843fa#diff-3bccf87f2d5228b9b7bc11c1ad0e8656R1042); rolling back to 0.3.1.5 fixes the issue. Possible long-term fix would be using null-safe [`Objects.equals()`](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#equals-java.lang.Object-java.lang.Object-).

Relevant [error.log](https://github.com/Innoxia/liliths-throne-public/files/2970143/error.log)

PR #1105, opened 2019-03-15
Bugfix: Make imp fortress subspecies checks null-safe

Niranufoti:
This is a fix for issue #1104. See that issue for details.

Tested against 0.3.1.6.


Innoxia:
Thank you!

PR #1108, opened 2019-03-16
Fixes some flags in SlaveDialogue

GynoidDoesGames:
Before these commits, consensuality in these scenes was being ignored/discarded after being calculated, always returning "false". These changes fix the scenes to properly take the calculations into account.


Innoxia:
Thank you!

PR #1110, opened 2019-03-16
Add difficulty option: Alternative Forced TF

0d721077:

*This adds one of some proposed optional difficulty/scenario options.*

*This difficulty option will make the potions of random encounter victors much more potent and varying while still respecting the victors preferences.*
  - *Potions will (in almost every case) additionally grant a fetish or alter one desire*
  - *Victors do not need the Kink Advocate or Transformer Fetish to try forcing the potion on you*
  - *Potions won't focus solely on genitals when preference is not matched*
  - *overall potency of potions varies with kinks of the victor*

*no*

*by me, with various versions from 3.0+*

*none currently*

Since i'm not a native english speaker, the wording may seem a bit odd at times and i'll appreciate corrections/suggestions!

### DIFF
--- a/src/com/lilithsthrone/controller/MainControllerInitMethod.java
+++ b/src/com/lilithsthrone/controller/MainControllerInitMethod.java
@@ -5028,7 +5028,9 @@ public class MainControllerInitMethod {
 		
 		// Content preferences:
 
-		if (Main.game.getCurrentDialogueNode() == OptionsDialogue.CONTENT_PREFERENCE || Main.game.getCurrentDialogueNode() == CharacterCreation.CONTENT_PREFERENCES) {
+		if (Main.game.getCurrentDialogueNode() == OptionsDialogue.CONTENT_PREFERENCE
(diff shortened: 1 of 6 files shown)


Eliria:
Looks like it works! 
Tested this some. I lowered the strength of its effects since it can give a lot. It adds a much faster TF pace (and varied TFs that don't always start with genitals) which I appreciate if I want to explore a form from an npc without submitting dozens of times. I'm not sure the implementation fits as is but I like the direction.

0d721077:
Glad you like it. I'm unsure about the overall strength of the potions aswell, but as random encounter fights become easy to win very quick i thought there had to be a massive penalty for losing. And getting back from submission as Croc-Rat-Bat-abomination with a few new fetishes seemed to be a fitting penalty for getting in too deep :)

Since i first implemented it pre 3.0 there are no taur-options yet. Sadly (or luckily?), i currently lack the time to do that but i guess the new TFPotionGenerator class could be easily extended by anyone.

PR #1113, opened 2019-03-16
Add difficulty option: Opportunistic Attackers

0d721077:

*This adds one of some proposed optional difficulty/scenario options.*

*This difficulty option will increase the probability of random encounters when the player is beat up or obviously easy prey. The increased probability is only applied to encounters that want to take advantage of the player. The encounter probability "boost" is determined by:*
  - *energy level*
  - *lust level*
  - *dirty clothing/body*
  - *exposed body parts*
  - *"drunk" state*

*no*

*by me, with various versions from 3.0+*

*none currently*

Since i'm not a native english speaker, the wording may seem a bit odd at times and i'll appreciate corrections/suggestions!

### DIFF
--- a/src/com/lilithsthrone/controller/MainControllerInitMethod.java
+++ b/src/com/lilithsthrone/controller/MainControllerInitMethod.java
@@ -5028,7 +5028,9 @@ public class MainControllerInitMethod {
 		
 		// Content preferences:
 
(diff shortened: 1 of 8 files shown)


Innoxia:
Thank you for this!

PR #1121, opened 2019-04-03
Birthing type check based on Boolean value and Slime Breast TF Restore

DL-Vadric:
To modify the properties for live births versus laying eggs so that they can be applied without relying on another switch statement. I saw there was a prior pull request from December doing something similar but, at least from the code, it appeared as though the intent was to base this on the vagina type.

Added a Boolean property to VaginaType.java indicating if it is an egg-laying type or a live birth type. Changed Lilaya's lab dialogue to refer to the boolean in lieu of a case value.

No

Yes, it has been tested on v0.3.2 using human, wolf, alligator, and harpy types.

Vadric#5183


src/com/lilithsthrone/game/character/body/types/VaginaType.java
```diff
@@ -17,46 +17,47 @@
...
 	private FluidType fluidType;
 	private Race race;
+	private boolean eggLaying;
 	private List<OrificeModifier> defaultRacialOrificeModifiers;
 
-	private VaginaType(BodyCoveringType skinType, FluidType fluidType, Race race, OrificeModifier... defaultRacialOrificeModifiers) {
+	private VaginaType(BodyCoveringType skinType, FluidType fluidType, Race race, boolean eggLaying, OrificeModifier... defaultRacialOrificeModifiers) {
 		this.skinType = skinType;
 		this.fluidType = fluidType;
 		this.race = race;
```
Stadler76: I assume
```java
		this.eggLaying = eggLaying;
```
is missing here?

DL-Vadric: I don't even know how I got the commit ahead of fixing that, but it's obvious that I did. Thank you for catching that. Fixed now.

DL-Vadric:
To be reviewed.

PR #1124, opened 2019-04-12
Fix virginity cost errors

GynoidDoesGames:
Fixes slavery value for characters w/ retained virginity. Also adds an underflow check in there for good measure.

@Innoxia

### DIFF
--- a/src/com/lilithsthrone/game/character/GameCharacter.java
+++ b/src/com/lilithsthrone/game/character/GameCharacter.java
@@ -3653,12 +3653,7 @@ public abstract class GameCharacter implements XMLSaving {
 			}
 		}
 		
-		if(value < 0)
-		{
-			value = 0;
-		}
-		
-		return value;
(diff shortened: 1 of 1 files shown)


src/com/lilithsthrone/game/character/GameCharacter.java
```diff
@@ -3653,6 +3653,11 @@ public int getValueAsSlave(boolean includeInventory) {
 			}
 		}
 		
+		if(value < 0)
+		{
+			value = 0;
+		}
+		
 		return value;
```
Pimvgd: return Math.max(0, value);

GearheadLydia: Thanks.

GynoidDoesGames: Thanks.

Innoxia:
Thank you!

PR #1125, opened 2019-04-13
Deprecated Method Used

King-Drake:
The method BigDecimal.setScale(int, int) is deprecated in favor of BigDecimal.setScale(int, RoundingMode)

Mostly I'm learning to use GitHub, focusing on minor fixes.
Replacing deprecated code with the up-to-date version.
The deprecated code will still compile in Java 12, but gives a deprecation warning in Java 10 or higher.  The fix is compatible starting in Java 5, and should be functional indefinitely.  Since Lilith's Throne is compiled in Java 8 to avoid issues with JavaFX, this should not be a major concern.
JJoseph#3701

### DIFF
--- a/src/com/lilithsthrone/utils/Units.java
+++ b/src/com/lilithsthrone/utils/Units.java
@@ -574,7 +574,7 @@ public enum Units {
      */
     public static float round(double value, int places) {
         if (places < 0) throw new IllegalArgumentException("Amount of fractional places cannot be less than 0.");
-        return BigDecimal.valueOf(value).setScale(places, java.math.RoundingMode.HALF_UP).floatValue();
+        return BigDecimal.valueOf(value).setScale(places, RoundingMode.HALF_UP).floatValue();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/com/lilithsthrone/utils/Units.java
```diff
@@ -574,7 +574,7 @@ public static String getMonthName(int month) {
      */
     public static float round(double value, int places) {
         if (places < 0) throw new IllegalArgumentException("Amount of fractional places cannot be less than 0.");
-        return BigDecimal.valueOf(value).setScale(places, BigDecimal.ROUND_HALF_UP).floatValue();
+        return BigDecimal.valueOf(value).setScale(places, java.math.RoundingMode.HALF_UP).floatValue();
```
Pimvgd: RoundingMode is already on the import list, so replace with RoundingMode.HALF_UP (without the java.math. prefix)

Innoxia: Applied the suggested change

Innoxia:
There was some extra baggage when changing branch to dev (I don't like committing PRs directly to master). I've applied this change locally and will credit you for it. Thank you.

Issue #1127, opened 2019-04-14
Incomplete demon TF, when TFing from a taur-body


Stadler76:
When I go through the '**Become a demon**' event starting as a taur (and probably any other non-biped leg-configuration) my legs stay unchanged. Obviously, because `LegType.DEMON_COMMON` is not allowed for demon-taurs.

In addition to that, my crotch-boobs stay unchanged, too.

A possible solution would be changing the `legType` to `LegType.DEMON_HOOFED` for taurs and change the `LegConfiguration` to `BIPEDAL` for any other non-bipeds. And to change the type of the crotch boobs, if the player has any.

Untested, but something, like this should work:
```java
switch (Main.game.getPlayer().getLegConfiguration()) {
	case BIPEDAL:
		Main.game.getPlayer().setLegType(LegType.DEMON_COMMON);
		break;

	case TAUR:
		Main.game.getPlayer().setLegType(LegType.DEMON_HOOFED);
		break;

	default:
		Main.game.getPlayer().setLegConfiguration(LegConfiguration.BIPEDAL);
		Main.game.getPlayer().setLegType(LegType.DEMON_COMMON);
}
```

Stadler76:
Ok, I've tested it while opening this issue. Possible fix: [20190414_2147_demon-taur-fix.diff.txt](https://github.com/Innoxia/liliths-throne-public/files/3077813/20190414_2147_demon-taur-fix.diff.txt)
Note, that the comments about `setBreastCrotchSize` in that diff are hints for you, since I couldn't decide, if to set them to a fixed value or to leave them as is.

Zsar:
Code as written does not actually require the `case BIPEDAL`.

I'd say `default` ought to throw an exception though, instead of quasi-randomly mapping to another case. This way there is actually a chance for this issue to be detected on introduction of a new leg configuration.

… Might want to then write a Unit Test for the method that iterates through `LegConfiguration.values()` - that would then be guaranteed to crash before release and can be noticed and fixed before ever going public.

The method also ought to take the character as parameter, so it works for all characters, not just the PC.

Stadler76:
@Zsar wrote:
> Code as written does not actually require the case BIPEDAL.
> 
> I'd say default ought to throw an exception though, instead of quasi-randomly mapping to another case. This way there is actually a chance for this issue to be detected on introduction of a new leg configuration.

[...]

PR #1128, opened 2019-04-14
Fix taur to demon TF

Stadler76:
  This is the proposed fix for issue #1127

  'Become a demon' didn't take taur body into consideration. In addition to that I've made it, that the wing size is set to the minimum requirement to fly for the leg configuration.

  No

  Yes, 0.3.2.2 Alpha. Gone through 'Become a demon' and turning Lilaya and Meraxis into demons.

  Stadler#3007

### DIFF
--- a/src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
+++ b/src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
@@ -74,7 +74,28 @@ public class SALyssiethSpecials {
 		Main.game.getPlayer().setPenisCumExpulsion(FluidExpulsion.FOUR_HUGE.getMedianValue());
 		Main.game.getPlayer().fillCumToMaxStorage();
 	}
-	
+
+	public static final void playerGrowDemonicBreasts(CupSize breastCup, CupSize crotchBreastCup) {
+		if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
+			Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
+			if (crotchBreastCup!=null) {
+				Main.game.getPlayer().setBreastCrotchSize(crotchBreastCup);
+			}
+		}
+		Main.game.getPlayer().setBreastType(BreastType.DEMON_COMMON);
+		if(breastCup!=null) {
+			Main.game.getPlayer().setBreastSize(breastCup);
+		}
+	}
+
+	public static final void playerGrowDemonicBreasts(CupSize breastCup) {
+		playerGrowDemonicBreasts(breastCup, null);
+	}
+
+	public static final void playerGrowDemonicBreasts() {
+		playerGrowDemonicBreasts(null, null);
+	}
+
 	public static final SexAction DEMON_TF_STAGE_1_CHOICE_GET_FUCKED = new SexAction(
 			SexActionType.SPECIAL,
 			ArousalIncrease.ONE_MINIMUM,
@@ -132,12 +153,7 @@ public class SALyssiethSpecials {
 			Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 			Main.game.getPlayer().setHeight(168);
 			Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
-			if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
-				Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
-				// Main.game.getPlayer().setBreastCrotchSize(CupSize.C); // leave as is?
-			}
-			Main.game.getPlayer().setBreastType(BreastType.DEMON_COMMON);
-			Main.game.getPlayer().setBreastSize(CupSize.DD);
+			playerGrowDemonicBreasts(CupSize.DD, CupSize.C);
 			Main.game.getPlayer().setPenisType(PenisType.NONE);
 			playerGrowDemonicVagina();
 			
@@ -205,13 +221,7 @@ public class SALyssiethSpecials {
 				Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 				Main.game.getPlayer().setHeight(168);
 				Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
-				if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
-					Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
-					// Main.game.getPlayer().setBreastCrotchSize(CupSize.TRAINING_AA); // leave as is?
-				}
-				Main.game.getPlayer().setBreastType(BreastType.DEMON_COMMON);
-				Main.game.getPlayer().setBreastSize(CupSize.AA);
-				
+				playerGrowDemonicBreasts(CupSize.AA, CupSize.TRAINING_AA);
 			} else {
 				Main.game.getPlayer().incrementFemininity(15);
 				Main.game.getPlayer().setHairLength(HairLength.FOUR_MID_BACK.getMedianValue());
@@ -221,12 +231,7 @@ public class SALyssiethSpecials {
 				Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 				Main.game.getPlayer().setHeight(168);
 				Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
-				if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
-					Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
-					// Main.game.getPlayer().setBreastCrotchSize(CupSize.C); // leave as is?
-				}
-				Main.game.getPlayer().setBreastType(BreastType.DEMON_COMMON);
-				Main.game.getPlayer().setBreastSize(CupSize.DD);
+				playerGrowDemonicBreasts(CupSize.DD, CupSize.C);
 			}
 			
 			if(Main.game.getPlayer().hasPenis()) {
@@ -298,10 +303,7 @@ public class SALyssiethSpecials {
 			
 			playerGrowDemonicPenis();
 			Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
-			if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
-			    Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
-			}
-			Main.game.getPlayer().setBreastType(BreastType.DEMON_COMMON);
+			playerGrowDemonicBreasts();
 			
 			if(!Main.game.getPlayer().isFeminine()) {
 				Main.game.getPlayer().setHeight(208);
@@ -637,12 +639,7 @@ public class SALyssiethSpecials {
 			Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 			Main.game.getPlayer().setHeight(168);
 			Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
-			if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
-			    Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
-			    // Main.game.getPlayer().setBreastCrotchSize(CupSize.DD); // leave as is?
-			}
-			Main.game.getPlayer().setBreastType(BreastType.DEMON_COMMON);
-			Main.game.getPlayer().setBreastSize(CupSize.F);
+			playerGrowDemonicBreasts(CupSize.F, CupSize.DD);
 			playerGrowDemonicVagina();
 			
 			Sex.setSexManager(
@@ -708,12 +705,7 @@ public class SALyssiethSpecials {
 				Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 				Main.game.getPlayer().setHeight(168);
 				Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
-				if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
-					Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
-					// Main.game.getPlayer().setBreastCrotchSize(CupSize.DD); // leave as is?
-				}
-				Main.game.getPlayer().setBreastType(BreastType.DEMON_COMMON);
-				Main.game.getPlayer().setBreastSize(CupSize.F);
+				playerGrowDemonicBreasts(CupSize.F, CupSize.DD);
 				
 			} else {
 				Main.game.getPlayer().setFemininity(50);
@@ -724,12 +716,7 @@ public class SALyssiethSpecials {
 				Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 				Main.game.getPlayer().setHeight(168);
 				Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
-				if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
-					Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
-					// Main.game.getPlayer().setBreastCrotchSize(CupSize.TRAINING_AA); // leave as is?
-				}
-				Main.game.getPlayer().setBreastType(BreastType.DEMON_COMMON);
-				Main.game.getPlayer().setBreastSize(CupSize.AA);
+				playerGrowDemonicBreasts(CupSize.AA, CupSize.TRAINING_AA);
 			}
 			
 			Sex.setSexManager(
--- a/src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
+++ b/src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
@@ -134,7 +134,7 @@ public class SALyssiethSpecials {
 			Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
 			if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
 				Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
-				// Main.game.getPlayer().setBreastCrotchSize(CupSize.D); // leave as is?
+				// Main.game.getPlayer().setBreastCrotchSize(CupSize.C); // leave as is?
 			}
 			Main.game.getPlayer().setBreastType(BreastType.DEMON_COMMON);
 			Main.game.getPlayer().setBreastSize(CupSize.DD);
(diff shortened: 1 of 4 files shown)


src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -132,6 +132,10 @@ public void applyEffects() {
 			Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 			Main.game.getPlayer().setHeight(168);
 			Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
+			if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
+				Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
+				// Main.game.getPlayer().setBreastCrotchSize(CupSize.D); // leave as is?
```
Stadler76: Please tell me, if this should be removed aka leave the crotch cup size as is or to be set to a certain value. Personally I'd prefer the former.

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -201,6 +205,10 @@ public void applyEffects() {
 				Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 				Main.game.getPlayer().setHeight(168);
 				Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
+				if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
+					Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
+					// Main.game.getPlayer().setBreastCrotchSize(CupSize.D); // leave as is?
```
Stadler76: same as above

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -213,6 +221,10 @@ public void applyEffects() {
...
```
Stadler76: same as above

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -622,6 +637,10 @@ public void applyEffects() {
 			Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 			Main.game.getPlayer().setHeight(168);
 			Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
+			if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
+			    Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
+			    // Main.game.getPlayer().setBreastCrotchSize(CupSize.D); // leave as is?
```
Stadler76: same as above

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -689,6 +708,10 @@ public void applyEffects() {
 				Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 				Main.game.getPlayer().setHeight(168);
 				Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
+				if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
+					Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
+					// Main.game.getPlayer().setBreastCrotchSize(CupSize.D); // leave as is?
```
Stadler76: same as above

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -701,6 +724,10 @@ public void applyEffects() {
...
```
Stadler76: same as above

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -132,6 +132,10 @@ public void applyEffects() {
 			Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 			Main.game.getPlayer().setHeight(168);
 			Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
+			if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
+				Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
+				// Main.game.getPlayer().setBreastCrotchSize(CupSize.C); // leave as is?
```
Stadler76: Please tell me, if this should be removed aka leave the crotch cup size as is or to be set to a certain value. Personally I'd prefer the former.

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -201,6 +205,10 @@ public void applyEffects() {
 				Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 				Main.game.getPlayer().setHeight(168);
 				Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
+				if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
+					Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
+					// Main.game.getPlayer().setBreastCrotchSize(CupSize.TRAINING_AA); // leave as is?
```
Stadler76: same as above

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -213,6 +221,10 @@ public void applyEffects() {
 				Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 				Main.game.getPlayer().setHeight(168);
 				Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
+				if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
+					Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
+					// Main.game.getPlayer().setBreastCrotchSize(CupSize.C); // leave as is?
```
Stadler76: same as above

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -622,6 +637,10 @@ public void applyEffects() {
 			Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 			Main.game.getPlayer().setHeight(168);
 			Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
+			if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
+			    Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
+			    // Main.game.getPlayer().setBreastCrotchSize(CupSize.DD); // leave as is?
```
Stadler76: same as above

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -689,6 +708,10 @@ public void applyEffects() {
 				Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 				Main.game.getPlayer().setHeight(168);
 				Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
+				if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
+					Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
+					// Main.game.getPlayer().setBreastCrotchSize(CupSize.DD); // leave as is?
```
Stadler76: same as above

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -701,6 +724,10 @@ public void applyEffects() {
 				Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 				Main.game.getPlayer().setHeight(168);
 				Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
+				if(Main.game.getPlayer().getBreastCrotchType()!=BreastType.NONE) {
+					Main.game.getPlayer().setBreastCrotchType(BreastType.DEMON_COMMON);
+					// Main.game.getPlayer().setBreastCrotchSize(CupSize.TRAINING_AA); // leave as is?
```
Stadler76: same as above

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -132,8 +153,7 @@ public void applyEffects() {
 			Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 			Main.game.getPlayer().setHeight(168);
 			Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
-			Main.game.getPlayer().setBreastType(BreastType.DEMON_COMMON);
-			Main.game.getPlayer().setBreastSize(CupSize.DD);
+			playerGrowDemonicBreasts(CupSize.DD, CupSize.C);
```
Stadler76: Please tell me, if only the breast cup size should be set aka leave the crotch cup size as is or to be set to a certain value, like I did here. Personally I'd prefer the former. And if the latter: What `CupSize` should the crotch boobs actually get?

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -201,9 +221,7 @@ public void applyEffects() {
 				Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 				Main.game.getPlayer().setHeight(168);
 				Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
-				Main.game.getPlayer().setBreastType(BreastType.DEMON_COMMON);
-				Main.game.getPlayer().setBreastSize(CupSize.AA);
-				
+				playerGrowDemonicBreasts(CupSize.AA, CupSize.TRAINING_AA);
```
Stadler76: same as above

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -213,8 +231,7 @@ public void applyEffects() {
 				Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 				Main.game.getPlayer().setHeight(168);
 				Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
-				Main.game.getPlayer().setBreastType(BreastType.DEMON_COMMON);
-				Main.game.getPlayer().setBreastSize(CupSize.DD);
+				playerGrowDemonicBreasts(CupSize.DD, CupSize.C);
```
Stadler76: same as above

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -622,8 +639,7 @@ public void applyEffects() {
 			Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 			Main.game.getPlayer().setHeight(168);
 			Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
-			Main.game.getPlayer().setBreastType(BreastType.DEMON_COMMON);
-			Main.game.getPlayer().setBreastSize(CupSize.F);
+			playerGrowDemonicBreasts(CupSize.F, CupSize.DD);
```
Stadler76: same as above

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -689,8 +705,7 @@ public void applyEffects() {
 				Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 				Main.game.getPlayer().setHeight(168);
 				Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
-				Main.game.getPlayer().setBreastType(BreastType.DEMON_COMMON);
-				Main.game.getPlayer().setBreastSize(CupSize.F);
+				playerGrowDemonicBreasts(CupSize.F, CupSize.DD);
```
Stadler76: same as above

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
```diff
@@ -701,8 +716,7 @@ public void applyEffects() {
 				Main.game.getPlayer().setBodySize(BodySize.TWO_AVERAGE.getMedianValue());
 				Main.game.getPlayer().setHeight(168);
 				Main.game.getPlayer().setAssType(AssType.DEMON_COMMON);
-				Main.game.getPlayer().setBreastType(BreastType.DEMON_COMMON);
-				Main.game.getPlayer().setBreastSize(CupSize.AA);
+				playerGrowDemonicBreasts(CupSize.AA, CupSize.TRAINING_AA);
```
Stadler76: same as above

Zsar:
([start of discussion](https://github.com/Innoxia/liliths-throne-public/issues/1127#issuecomment-483435846))

Mmh. Common Unit Test frameworks for Java are [JUnit](https://junit.org/) or [TestNG](https://testng.org/doc/index.html), but this project currently does not use either one (or indeed any test framework; at least test scope is available "natively" through Maven, if currently also unused).
I'll try to set something up over the weekend and get it Pull-Request-worthy.

It did occur to me, however, that there is a way to check for missing LegConfigurations at compile time:
1. turn LegConfiguration into an abstract class with protected constructor and abstract method transformToDemon (mayhap later to be refactored to take a transformation target)
2. for every one of the current enum entries, create a child class of LegConfiguration which implements the transformToDemon method
3. instead of the switch statement handling all the LegConfiguration instances (outside of their own class files, where it may easily be missed - as has already happened, evidenced by your issue #1127 ), just get the character's instance and call its transformToDemon method

Result: A new LegConfiguration will only compile when it can provide the appropriate demon transformation; one must consciously implement a placeholder (and ought to open an issue at the same time, so documentation exists for a later fix) to leave it out.
Even less likely to be missed than a Unit Test that may or may not actually be run and investigated!

… If I am not mistaken, this would be a "textbook" OOD approach to a "textbook" OOD code smell (this particular pattern of switch use).
A bit less elegant than it could be, because Java does not support enum subclassing, but oh well!

------

Unrelated to the above, I would think that this transformation "atomically" should not change body part sizes, should it? Considering that demons are shapechangers in-game, it seems odd that to become one (unlike turning into something with defined "median dimensions") would incur changes to, well, shape beyond the minimally defining feature (swapping body parts for "Demonic" versions, that is).
… As a caveat, I have not yet played through the corresponding vanilla scenes, so mayhap my reasoning has already been invalidated by "word of god"?

In any case, I'd expect most players to react to their dimensional changes by just changing back through their new-won powers (or transformative potions) - why bother them with this diversion?

Stadler76:
I could add the methods
```Java
public String setMinumumBreastSize(int size) {
	return setBreastSize(Math.max(getBreastRawSizeValue(), size));
}

public String setMinumumBreastSize(CupSize size) {
	return setBreastSize(Math.max(getBreastRawSizeValue(), size.getMeasurement()));
}
```
so that breast sizes can only get bigger, but not smaller, when turning demon, but I'm a bit reluctant to bloat [GameCharacter.java](https://github.com/Innoxia/liliths-throne-public/blob/master/src/com/lilithsthrone/game/character/GameCharacter.java) even more ...

Stadler76:
I could add the methods
```Java
public String setMinumumBreastSize(int size) {
	return setBreastSize(Math.max(getBreastRawSizeValue(), size));
}

public String setMinumumBreastSize(CupSize size) {
	return setBreastSize(Math.max(getBreastRawSizeValue(), size.getMeasurement()));
}
```
so that breast sizes can only get bigger but not smaller when turning demon, but I'm a bit reluctant to bloat [GameCharacter.java](https://github.com/Innoxia/liliths-throne-public/blob/master/src/com/lilithsthrone/game/character/GameCharacter.java) even more ...

Zsar:
I think that a relative change would be better than an absolute one, indeed. But why have size change at all? I fail to see the motivation.

For the scope of this fix however, I guess the PR is in a pretty good spot now.
Open questions remain, e.g. why so much transformation logic has to be reimplemented in `SALyssiethSpecials `and `Lyssieth` when the functionality to do this ought to be available for far longer than Lyssieth has existed in code, but in the interest of small, incremental changes, I guess we'd better not dwell on them for now.

It might be a good idea to move further breast changes into a new, separate branch. After all, they are certainly unrelated to the original issue. What you have now seems to capture the spirit of the original implementation quite well.

… There ought to be some functionality that already allows increasing body part size by at least a single step, bounds-safe: Recurring changes from magical clothing already have do that. I think it's `AbstractItemEffectType.getLimits` and `EnchantmentEventListener.handleEvent` that govern this.
Moving from there might be the better choice vs. duplicating.

Stadler76:
> I think that a relative change would be better than an absolute one, indeed. But why have size change at all? I fail to see the motivation.

Well. I guess to avoid succubi with tiny breasts? They always have big jugs. :-)

> For the scope of this fix however, I guess the PR is in a pretty good spot now.
> Open questions remain, e.g. why so much transformation logic has to be reimplemented in `SALyssiethSpecials `and `Lyssieth` when the functionality to do this ought to be available for far longer than Lyssieth has existed in code, but in the interest of small, incremental changes, I guess we'd better not dwell on them for now.
> 
> It might be a good idea to move further breast changes into a new, separate branch. After all, they are certainly unrelated to the original issue. What you have now seems to capture the spirit of the original implementation quite well.
> 
> … There ought to be some functionality that already allows increasing body part size by at least a single step, bounds-safe: Recurring changes from magical clothing already have do that. I think it's `AbstractItemEffectType.getLimits` and `EnchantmentEventListener.handleEvent` that govern this.
> Moving from there might be the better choice vs. duplicating.

Short answer: Probably some refactoring is needed mid to long term, but that would be too much for this PR, since this is to fix a bug. I guess that should be done in new PRs and/or issues.

Zsar:
I concur that this pull request is good to go.

… For what it's worth, I have just created pull request #1139 for TestNG integration offering Unit Test and Integration Test support. Figured it might be of interest.

src/com/lilithsthrone/game/sex/sexActions/submission/SALyssiethSpecials.java
Innoxia: I did as you suggested, and changed it so that crotch-breast size is left as it is in all instances.

Innoxia:
Thank you for this! I've merged it into my local code, which should show up here on github when I push later tonight. ^^

PR #1136, opened 2019-04-20
Performance fixes for savefiles

Pimvgd:
After noticing that loading a savefile could take up to 19 seconds, Bast and I checked with a profiler to find places with performance issues. The resulting fixes brought the loading times back down to 4 seconds. It probably also significantly affected game startup time.

- Fixes a typo in clothing file which means we don't have to run the string matcher in that case
- Fixes the string matcher to check for the best case first, which seems to apply in pretty much all cases after fixing the one typo I saw earlier.

That's one's a massive speed up.

I tried some other assorted attempts at speeding the game up, which may or may not really have worked. It at least seems sensible to not go printing how a character feels about their lust being loaded, but it's not something that you'd notice as much as not comparing all the strings everywhere.

### DIFF
--- a/src/com/lilithsthrone/game/character/GameCharacter.java
+++ b/src/com/lilithsthrone/game/character/GameCharacter.java
@@ -4032,12 +4032,7 @@ public abstract class GameCharacter implements XMLSaving {
 	}
 	
 	public boolean hasDiscoveredElemental() {
-		try {
-			Main.game.getNPCById(elementalID);
-			return true;
-		}catch(Exception e) {
-			return false;
-		}
+		return Main.game.isCharacterExisting(elementalID);
 	}
 	
 	public boolean isElementalSummoned() {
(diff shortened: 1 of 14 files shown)


src/com/lilithsthrone/game/character/GameCharacter.java
```diff
@@ -13431,6 +13417,18 @@ public String setLust(float lust) {
...
 	}
+
+	public void setLustNoText(float lust) {
+		if (lust < 0) {
+			setAttribute(Attribute.LUST, 0, false);
+		} else if (lust > 100) {
+			setAttribute(Attribute.LUST, 100, false);
+		} else {
+			setAttribute(Attribute.LUST, lust, false);
+		}
```
Stadler76: This *might* be shortened with:
```java
		setAttribute(Attribute.LUST, Math.max(0, Math.min(lust, 100)), false);
```

Pimvgd: You're right, all I did was move it around and so I didn't really have a good look, but less code is better

Innoxia:
Wow, that's a huge improvement, thank you so much!

PR #1137, opened 2019-04-20
Clean character inventory

Pimvgd:
Less code is more


Innoxia:
Thank you for this! It makes it a lot easier to read.

Issue #1133, opened 2019-04-19
Demons in the Streets


AnonDragonWolfess:
"The only people able to withstand the storm's thunderous power are demons, and every now and then you see one strutting down the street.​ They sometimes cast a curious glance your way, but most are content to simply ignore you.​"

This plays even after you've been transformed into a demon. It would make more sense if they were no longer be curious when you're in your demon form, and become curious again if you were to take on the appearance of a human again.

DL-Vadric:
Pull request #1141 created as a proposed edit for this issue.

Innoxia:
Fixed in #1141

PR #1141, opened 2019-04-22
Arcane storm background demon characters

DL-Vadric:
Addresses issue #1133 - Slightly modifies the description of background characters on the streets during an arcane storm if the player is a demon and defaults to the original text if not.


Innoxia:
Thank you for this! There was another instance in the same file that needed changing, and I slightly changed the code, but left your description intact. I'll credit you in the credits for this; thanks again!

PR #1145, opened 2019-04-28
Alter clothing blocked by message to not include the clothing which is blocked

Pimvgd:
This doesn't fix much, but it does fix the case where you'd be told that you can't unequip X as X is blocked by Y and X. It should now state you can't unequip X because it is blocked by Y.


Innoxia:
Thank you!

PR #1146, opened 2019-05-01
Character inventory cleaning

Pimvgd:
Requires merging #1145 first.

### DIFF
--- a/src/com/lilithsthrone/game/inventory/AbstractInventory.java
+++ b/src/com/lilithsthrone/game/inventory/AbstractInventory.java
@@ -1,0 +1,98 @@
+package com.lilithsthrone.game.inventory;
+
+import java.util.*;
+import java.util.function.Function;
+
+/**
+ * AbstractInventory serves to deduplicate code between the handling of Item, Clothing and Weapon in CharacterInventory.
+ *
+ * Package-privateness is a feature in that case.
+ * @param <T>
+ */
+class AbstractInventory<T extends AbstractCoreItem, U extends AbstractCoreType> {
+	private final Comparator<T> comparator;
+	private final Function<T, U> typeRetriever;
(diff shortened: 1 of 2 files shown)


src/com/lilithsthrone/game/inventory/AbstractInventory.java
```diff
@@ -0,0 +1,109 @@
...
+
+	boolean isEmpty() {
+		return duplicateCounts.isEmpty();
+	}
+
+	public void sort() {
+		if (duplicateCounts.size() < 2) {
+			return;
+		}
+		List<T> weaponsToSort = new ArrayList<>(duplicateCounts.keySet());
```
CognitiveMist: Minor gripe: if this is applicable to items, clothing, and weapons, then this variable shouldn't be called weaponsToSort.

Innoxia: I'll get that changed on my end after merging.

Innoxia:
Thank you!

Innoxia:
I just tested this locally, and trying to identify clothing in your inventory via spending arcane essences throws a null pointer exception.

src/com/lilithsthrone/game/inventory/AbstractInventory.java
Pimvgd: Fixed

Pimvgd:
Fixed null pointer; was caused by trying to be too fancy with computeIfPresent ... it's the one bit of code I had trouble with because I wanted to make it smaller and I ended up overcompressing. The new version is a lot more readable.

Innoxia:
Thank you for this! I've merged it locally, and will push to github in about 12 hours, so should show as merged then. ^^

Issue #1153, opened 2019-05-10
Trying sex with a partner while they are on a party


SephVEVO:
So, when you have a partner (one that you find on the streets and you get them home) on your party and try to sex them on their room, the game softlocks with your partner and a clone that's (_somehow_) on your party watching you jerk yourself off for eternity until you load or something.

To Reproduce:
1. Have someone that you took out from an alley (The ones that have the "talk" option when defeated) to love you (worship status);
2. Add them to your party;
3. On the same partner, select the option to have normal sex with them (note that the option "Side-by-Side" is available, even though it's only you and your partner that's on the room);
4. Enjoy your fapping session for eternity.

Additional information:
 - Game version: 0.3.2.3 alpha
 - OS: Windows 7
 - EXE Version.

A-haron:
Could reproduce @ 0.3.2.3 alpha prebuild (.exe)

This is what I got from the code:
[OccupantDialogue.OCCUPANT_START.getResponse(0,5)](https://github.com/Innoxia/liliths-throne-public/blob/ab24f7ea537b5b2523255392b520658c2d4175e3/src/com/lilithsthrone/game/dialogue/npcDialogue/OccupantDialogue.java#L219) and [GameCharacter.addCompanion()](https://github.com/Innoxia/liliths-throne-public/blob/ab24f7ea537b5b2523255392b520658c2d4175e3/src/com/lilithsthrone/game/character/GameCharacter.java#L4053) add the occupant as partner, as expected.

But, if you don't return to the room's menu, and use the first button on the second tab (Sex options), [OccupantDialogue.OCCUPANT_START.getResponse(1,1)](https://github.com/Innoxia/liliths-throne-public/blob/ab24f7ea537b5b2523255392b520658c2d4175e3/src/com/lilithsthrone/game/dialogue/npcDialogue/OccupantDialogue.java#L219). Creates a [SMGeneric](https://github.com/Innoxia/liliths-throne-public/blob/ab24f7ea537b5b2523255392b520658c2d4175e3/src/com/lilithsthrone/game/sex/managers/universal/SMGeneric.java#L33) with the occupant as submisive and dominant observer at the same time, which probably produces the error down the line.

Other options end up with the partner as Dominant and Submisive at the same time, or don't work at all (If you try putting "your partner" and "the occupant" as submisive at the same time).

Easiest fix is probabily to "reload the room" after adding an occupant to the party, as the occupant iteraction menu is disabled if the occupant is in your party.

[...]

PR #1154, opened 2019-05-11
Fix issue #1153

A-haron:
Fix issue #1153 

Return to the room after adding a occupant as party member, preventing access to options now unavailable.

Changed the nextDialogue from "OCCUPANT_START" to "Main.game.getDefaultDialogueNoEncounter()" after adding an occupant to the party.

~~**This change have not been tested yet, I haven't created the build environment yet.**~~
**Edit: Already tested, see [below](https://github.com/Innoxia/liliths-throne-public/pull/1154#issuecomment-491531990).**

If you want to test if this fixes the issue:

- Have an occupant able to join your party.
- Talk to it.
- Add it to the party.
- You shouldn't be able to keep talking in the occupant menu, as it's now disabled because "this character is currently working", effectively keeping you from initiating sex from the character's occupant menu, which produced the bug.

Going to test it later, once I have the build environment ready.

### DIFF
--- a/src/com/lilithsthrone/game/dialogue/npcDialogue/OccupantDialogue.java
+++ b/src/com/lilithsthrone/game/dialogue/npcDialogue/OccupantDialogue.java
@@ -218,7 +218,7 @@ public class OccupantDialogue {
 						} else if(Main.game.getPlayer().canHaveMoreCompanions()) {
 							return new Response("Add to party",
 									UtilText.parse(occupant(), "Ask [npc.name] if [npc.she] would like to accompany you for a while."),
-									OCCUPANT_START){
+									Main.game.getDefaultDialogueNoEncounter()){
 								@Override
 								public void effects() {
 									applyReactionReset();


SephVEVO:
I can't test it today but probably tomorrow. Lemme know what you find.

Thanks in advance by the way!

A-haron:
Just tested it out. Seems to work as intended, returning to the room after adding someone to the party.

The change is heavily scooped, there shouldn't be any side effects anywhere else.

As a longer proyect maybe adding a customised "I'll join" scene after asking to join may be a more elegant solution than returning to the room (With this pull) or returning to the initial scene (0.3.2.3 current way), but that's beyond the scope of the fix.

Innoxia:
Thank you for this! I ended up changing the fix slightly to make it so that sex options work even with room occupants being in your party, but I've still credited you in the patch notes for this. It will show up as merged when I push my local code to github later tonight.

PR #1156, opened 2019-05-12
Duplicate characters on ResponseSex

A-haron:
Prevent duplicate characters on sex scenes. At least those that pass through ResponseSex.

Added a check at ResponseSex constructor that deletes spectators that are involved in the sex scene itself.

This caused some errors, commonly with party members. [Example](https://github.com/Innoxia/liliths-throne-public/blob/ab24f7ea537b5b2523255392b520658c2d4175e3/src/com/lilithsthrone/game/dialogue/places/submission/impFortress/ImpFortressDialogue.java#L3016) (Using the main companion in the scene, but adding all companions (Main included) as spectators).

Nope.

Yes, 0.3.2.3 dev branch.

A-haron#3467

### DIFF
--- a/src/com/lilithsthrone/game/dialogue/responses/ResponseSex.java
+++ b/src/com/lilithsthrone/game/dialogue/responses/ResponseSex.java
@@ -394,6 +394,7 @@ public class ResponseSex extends Response {
 			}
 		}
 		
+		// Prevent duplicate characters on differente positions.
 		if(this.sexManager != null) {
 			
 			// We have to use iterator to be able to safely remove elements without "ConcurrentModificationException"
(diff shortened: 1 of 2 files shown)


A-haron:
Fix was moved from ResponseSex class to Sex class, as some scenes skipped the code in ResponseSex.

In the new position, the code prevents bugged scenes, but seems to cause NullPointerExceptions instead on those cases (With the scene not starting at all).

Adding some logging when a clone is removed to fix those scenes more easily is a possible improvement.

A-haron:
Talked with Innoxia, and this fix would hide more errors than it fixes, so it won't be implemented, at least this way. Closing.

Issue #1159, opened 2019-05-21
Unicorn Horn Issues


AnonDragonWolfess:
**Bug Description**
Growing unicorn horns via TF potion will NOT reset horns per row to 1 at the moment of transformation, as was stated in recent update. Not only this, but the unicorn horn does not register, making it impossible to become a unicorn- or alicorn-morph, or to morph someone into one. Unicorn- and alicorn-morphs can be spawned and registered properly in the Watering Hole, not sure if the same can be said for the back alleys or in the streets during an arcane storm.

**To Reproduce**
1. Make unicorn- or alicorn-morph transformative
2. Use on character (Can also use Transformation section in Buggy screen) or random encounter
3. See error

Version 0.3.2.6 Alpha

Stadler76:
According to this chunk of code:
```java
if(this.type.equals(HornType.NONE)) {
	UtilText.transformationContentSB.append(UtilText.parse(owner, 
			"<p>"
				+ "[npc.Name] [npc.verb(let)] out a surprised gasp and [npc.verb(rub)] at [npc.her] forehead as [npc.she] [npc.verb(feel)] it growing hot and sensitive."
				+ " After just a moment, [npc.her] [npc.eyes] widen in shock as something starts pushing out from under the [npc.faceSkin] of [npc.her] forehead."));
} else {
	if(type==HornType.HORSE_STRAIGHT) {
		this.setHornsPerRow(owner, 1);
	}
	UtilText.transformationContentSB.append(UtilText.parse(owner, 
			"<p>"
				+ "[npc.Name] [npc.verb(let)] out a surprised gasp as [npc.she] [npc.verb(feel)] an odd tingling sensation at the base of [npc.her] "+(owner.getTotalHorns()==1?"[npc.horn]":"[npc.horns]")+"."
				+ " Before [npc.she] [npc.has] any time in which to react, "+(owner.getTotalHorns()==1?"it rapidly crumbles  away, and within moments it's":"they rapidly crumble away, and within moments they've")+" completely disappeared. "));
}
```
... horns per row is only set to 1 if the character already has horns, but not, if the character has no horns at all.
@Innoxia: Was that intended?

Stadler76:
Oh, and now that I see it: Shouldn't that chunk also include ...
```java
this.setHornRows(owner, 1);
```
... since it requires to have a single horn, meaning one row with one horn, to be considered unicorn?

AnonDragonWolfess:
[...]

PR #1160, opened 2019-05-21
Fix unicorn horn issue

Stadler76:
  This is the proposed fix for issue #1159

  Horns per row were only being set, if the character already has horns. This will now be set, even if the character has no horns. In addition I've added, that the horn rows are set to 1 for unicorn horns.

  No

  Yes, 0.3.2.6 Alpha.

  Stadler#3007


Innoxia:
Thank you!

PR #1165, opened 2019-05-30
Fix appearance stating, that slimes can fly

Stadler76:
  Fix appearance stating for slimes, that their wings allow flight.

  The wing appearance chunk in `Body.java` didn't check for slimes. Fixed that.

  No

  Yes, 0.3.3 Alpha.

  Stadler#3007


Innoxia:
Thank you!

PR #1172, opened 2019-06-11
Fixed npc2 reference

TheKingInYellow:
Fix the display of error text in the incubus's main attack.

Filled out the target argument in UtilText.parse() DominionSuccubusAttacker

No

tested on latest dev - 3.3.6

TheKingInYellow


Innoxia:
Thank you for this!

Issue #1170, opened 2019-06-04
Null reference exception / Action buttons disappearing


Sakedo:
With corruption bypass disabled (not sure if that's required or not), using "Unrestrict position" on a defeated enemy results on a NullPointerException because the valid sex actions filters out all preferred moves that are not in the current position - and then it doesn't seem to be recomputed at any point, resulting in 0 available preferences.  

The symptom on the user's side is that all of the sex action buttons disappear (not grey out, literally vanish). 

The fix that I have working locally is to add in GenericActions.java / applyEffect() of both restrict position and permit position: 
((NPC)Sex.getCharacterTargetedForSexAction(this)).generateSexChoices(false, Main.game.getPlayer());

I'm confused why the preferred npc action list was not refreshing on its own, which makes me wonder if the list is being culled of anything not valid in the initial standing/standing even after the position is changed by the player.

AceXP:
This issue was fixed some time ago, could you check it and close this issue?

Issue #1158, opened 2019-05-18
Misc. Action sex buttons not working


xPURExRain:
**Describe the bug**
Whenever I have sex with anyone, the buttons allowing them to control clothing, as well as the button to offer pussy, either simply do not work upon clicking them, or do click however dont actually save the state of button, IE it remains showing Unrestrict Clothing and the AI seems to continue to follow it as if they wern't allowed

**Expected behavior**
I figured the expected behavior was for the buttons to work as intended, aka offer pussy wil have your character say something about you wanting them to use your pussy and push them in that direction. And for unrestrict clothing to let them take off your clothes.

**To Reproduce**
Please include the steps that need to be taken in order to reproduce the bug.
e.g.:
1. Start any sex
2. Click buttons
3. ???
4. See error (Profit)

**Attach files**
Please either attach a copy of your error.log (in the LilithsThrone/data directory), or copy & paste the contents to the end of this bug report.

**Additional information:**
 - Game version [e.g. 0.3.2.5]
 - OS: Windows 10
 - Using .jar
-
[Alot Happened.zip](https://github.com/Innoxia/liliths-throne-public/files/3194491/Alot.Happened.zip)

 If using .jar, and you are not including an error.log file/text, please include your java version: [e.g. 1.8.0_172]

AceXP:
This was fixed some time ago, could you check it and close this issue?

PR #1173, opened 2019-06-11
Solve NullReference exception with permit positioning

TheKingInYellow:
Fix issue  #1170  and #1158 

Invoked generateSexChoices after Permit or Restrict positioning is use to fix offending null reference exception

No

Yes, 3.3.6

TheKingInYellow


Stadler76:
> * What is the purpose of the pull request?
>   Fix issue  1170 and 1158

I'd suggest to add a `#` before the issue number so they get properly referenced.

Innoxia:
I'm pretty sure I've already fixed this bug on my end (or at least, I fixed **a** bug that was throwing NullPointerExceptions when unrestricting control/positioning), so I'm a little hesitant to merge this (as, by regenerating sex preferences, it has the potential to override any requested sex types that the player has made).

I won't close this just yet, in case it was a different bug that I fixed, but at the moment, I'm unable to reproduce the error that this PR fixes. If it's still an issue after the next public release, I'll take another look. ^^

PR #1175, opened 2019-06-13
Fix unintentional randomness in horn enchanting

Stadler76:
  If you have multiple horn types the enchantment UI randomized the selected horn type.

  The param `int index` wasn't being used in that corresponding function. Instead the `HornType` was chosen randomly. Changed that to actually use the index of the choice.

  No

  Yes, 0.3.3.6 Alpha.
  I haven't checked, if that breaks elsewhere aka, if the randomness was intended for 'other stuff', although I highly doubt that.

  Stadler#3007


Innoxia:
Thank you!

PR #1189, opened 2019-07-31
Flipped navel pentagram piercing

ItsNorin:
Someone revealed how fake of a goth I am for not knowing a star has to have a point down to be considered a pentagram.

Replacement .svg for piercing_pentagram_navel.svg


Rfpnj:
Well, I am going to add a bit of a different standpoint to the mix. The Point down is representative of a particular ideal. Most prefer the point up as it means the opposite of down in many ways. I am team Point up for one.

ItsNorin:
Huh didn't realize drafts are public lol
True, anyways asthetically I like the other current one more, just going to trash this.

PR #1190, opened 2019-08-13
Combat Multi-Target Effects Freeze/Crash Fix

Shadowheart329:
During combat, if a character uses an ability that has an effect that targets multiple combatants, it will fail thus freezing the turn when you end it. This is due to the fact that multi-target effects call the `getRandomAlliedPartyMember(GameCharacter target)` method. This method returns a random ally from the target's party. The problem is if the target is alone, the function still tries to find a random ally from the empty list, and since `random.nextInt()` can't accept a 0 value, it fails.

Basically, I just checked if the list of targeted allies was equal to 0, and if so just return the target itself, instead of trying to find a random ally from an empty list.

No.

Yes. Version 0.3.3.10

Shadowheart329#0896

### DIFF
--- a/src/com/lilithsthrone/game/combat/Combat.java
+++ b/src/com/lilithsthrone/game/combat/Combat.java
@@ -1543,6 +1543,9 @@ public enum Combat {
 				possibleTargets.add(character);
 			}
 		}
+		if(possibleTargets.size() == 0) {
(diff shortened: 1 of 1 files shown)


Innoxia:
Thank you for this!

PR #1193, opened 2019-08-24
fix health bug introduced in 03311

AlacoGit:

Fix a bug with health calculation that  was introduced in 7cb874c.
Added an additional check so the Aura penalty for having a low Arcane stat no longer also applies to Health.
No.
On the newest dev version(0.3.4).
Alaco on Discord.


Innoxia:
Thank you!

PR #1196, opened 2019-08-26
Potential copy error or typo in PlaceUpgrade.java

EPSIL0N:
Potential Typo / Copy Error
There was a reference to salve room double's list when defining slave room quadruple's list but that sort of cross reference is not seen elsewhere.

Discord tag GammaMike#6081


Innoxia:
Thank you, this was indeed a copy error!

PR #1200, opened 2019-08-26
[FIX] Changing position to Face sitting reversal caused exception

Sphilia:
[5:30 PM] Pilcrow: trying to set up reverse-facesitting puts an error in the log instead;
Error: getContent() method is throwing an exception in the node: 'Positioning Selection: Lying down'
and accepting the position generates no 'new position' text.

Renamed variables in a conditional from `faceSitting` to `faceSittingReversal`. Error possibly caused due to copy/paste.

No

0.3.4.1 Alpha

Sphilia : @Reyvateil#5830


Innoxia:
Thank you!

PR #1202, opened 2019-08-27
Allows slaves to gain xp from working

ItsNorin:
Allows slaves to gain a small amount of experiance from working. 
Chance to gain xp is proportional to how well they perform.

So we have a better idea of who you are, what is your Discord Handle?
Norin#3027


Innoxia:
This was merged as part of #1202

PR #1205, opened 2019-08-30
Threesome sex scene participant selection was inconsistantly substitu…

EPSIL0N:
Fix erroneous threesome participant selection

Changed some references from getslave or occupant to characterforsex or characterforsexsecondary in situations where the conditionals were already evaluating against characterforsex but the code for who to add to the sex scene would add getslave or occupant so you could end up with scenarios where you spoke to A who was not interested in sex with you, changed the targeting to B and C to enable the sex scene. The button text would mention A and C for example instead of B and C and then when you clicked start scene it would be A and C despite A not wanting to participate.

No

Tested against 3.4.1hotfix

 GammaMike 6081


Innoxia:
Thank you very much for this! I was meaning to go back over these files and fix my poorly-implemented code at some point, so this really saves me a lot of time - very much appreciated! ^^

PR #1210, opened 2019-09-04
Fixed Innoxia's super fun and interesting guide to parsing.

Miles-Light:
Now, the allmighty knowledge that you can obtain with Innoxia's super fun and interesting guide to parsing is even more amazing, because you can learn about the way targets work with Innoxia's super fun and interesting parser.


Miles-Light:
I actually didn't, i went a bit fast on that one. Sorry

PR #1213, opened 2019-09-12
Bugfix: Udders milk regeneration in transformation dialog

IjyiH:
Udders milk regeneration section in the demonic/slimey/elemental transformation dialog now points to udder milk regeneration instead of chest milk regeneration.
Tested on v0.3.4.6 without apparent issues.

@IjyiH#4260


RandomNecromancer:
I thought this might be happening, but I wasn't sure. Thanks for saving me the trouble of testing!

IjyiH:
Fixed in v0.3.5

PR #1227, opened 2019-11-04
Fix some stores (e.g. Clothing Emporium) being broken

triples941:
Added null check into PlayerCharacter.getQuest.
Without this patch some dialogues (e.g. Clothing Emporium, Arcane Arts) in v0.3.5.4 fail with NullPointerException.


Innoxia:
Thank you for this! I've edited my local code with your changes instead of directly merging this, due to my workflow being halfway through lots of other things, but I will credit you for the fix in the patch notes. ^^

Innoxia:
Actually nevermind, I am going crazy. I'll merge this now.

Issue #1228, opened 2019-11-04
Harpy encounters bugged: can't find text, and fires NullPointerException if lacking companion


thundahcoc:
**Describe the bug**
Been getting this message on harpy encounters:
> Dialogue for 'HARPY_ATTACK_[whatever]' not found! (Make sure that the 'res' folder is in the same directory as the .​jar or .​exe.​)

This is despite the _'res'_ folder indeed being there. No _"missing 'res' folder"_ popup on startup, and alleway encounters work perfectly fine.

Additionally, and if going alone, pressing any dialog button will get the following logged into the _'error.log'_:
```
Exception in thread "JavaFX Application Thread" java.lang.NullPointerException
	at com.lilithsthrone.game.dialogue.npcDialogue.dominion.HarpyAttackerDialogue.applyPregnancyReactions(HarpyAttackerDialogue.java:92)
	at ...
```
May be an easy fix: the conditional at [HarpyAttackerDialogue.java#L92](https://github.com/Innoxia/liliths-throne-public/blob/4c515a4a282ebc166b33b0636a9ece9587234fd3/src/com/lilithsthrone/game/dialogue/npcDialogue/dominion/HarpyAttackerDialogue.java#L92) just needs to check ```getMainCompanion()``` isn't ```null``` before calling ```.isVisiblyPregnant()```. 
However, I've no idea how that relates to the _'res'_ stuff, seeing as it happens even when the exception doesn't. Maybe some other missing ```null``` check?

**Expected behavior**
Text loading fine on harpy encounters, buttons working fine independently of wether I'm a lone or with a companion.

**To Reproduce**
Seems self explanatory, but they:
1. Start a new game
2. Make a beeline for the harpy nests
3. Explore until you encounter an harpy => Results in: "Dialogue for 'HARPY_ATTACK_[whatever]' not found!"
4. Press "Fight", "Offer Money" and/or "Offer Body" => Results in: "_NullPointerException_"

Additionally:
1. Start a new game/load a save
2. Obtain a companion (Friend, slave or elemental will do)
3. Make a beeline for the harpy nests
4. Explore until you encounter an harpy => Results in: "Dialogue for 'HARPY_ATTACK_[whatever]' not found!"
5. Press "Fight", "Offer Money", "Offer Body", "Offer Threesome", "Offer [companion]" => Game proceeds as normal

**Attach files**
[error.log](https://github.com/Innoxia/liliths-throne-public/files/3806267/error.log)

[...]

PR #1231, opened 2019-11-06
Fix attack dialogue problems

triples941:
Fix attack dialogue problems for Harpy Nests and Bat Caverns

Added check for companion for places where NPE was possible.
Changed Harpy Nests dialogue ID to match files.

No.

After the change I ran through affected areas in v0.3.5.4 again

~Have no Discord yet~ TripleS941#0897

P.S. Oops, got my non-lewd account involved. But it is not immediately obvious from there, so oh well.


triples941:
Basically a fix for all issues mentioned in #1228

Innoxia:
Thank you for this!

PR #1236, opened 2019-11-14
Alter race tooltip to better reflect knowledge from 'observant'

NoStepOnSneks:

Give players with the observant trait a quick idea of an NPC's general genital configuration without having to open the full description page to infer it from their true gender.
When the trait is active, instead of always showing "Penis: unknown, Vagina: unknown", the tool tip will now use a definitive "None" if the part is missing and "Unknown" if it's there but we haven't seen it in detail yet.
Also, the appropriate sentences on the full description page have been altered slightly to reflect that same certainty or doubt, for overall consistency.
no
tested in 0.3.5.4
No Step On Snek


Innoxia:
Thank you for this!

PR #1244, opened 2019-12-26
Fixing NullPointerExceptions when visiting Lyssieth

Stadler76:
Added a wrapper around getSubspeciesOverride().getRace() named getSubspeciesOverrideRace() which returns Race.NONE as a fallback. I simply wanted a fix, that avoids if (xyz() != null)-checks.


Stadler76:
wrong branch.
~closing~

PR #1248, opened 2019-12-28
Bat arms should block clothing

Stadler76:
Bat wings and harpy wings basically match their descriptions as being arms turned into wings. However: Bat wings don't block any clothing. This should fix that.

- Refactored `FITS_HARPY_WINGS` to `FITS_ARM_WINGS` -> meaning: fits any type of arm wings, whether it be harpy, bat or for example wyvern wings.
- Added `FITS_BAT_WINGS_EXCLUSIVE` and implemented it where applicable since I see that as for arm clothing that is made specifically for that race only. e. g. wyvern arms clothing only.

no

Version 0.3.5.8

Stadler#3007

### DIFF
--- a/src/com/lilithsthrone/game/character/GameCharacter.java
+++ b/src/com/lilithsthrone/game/character/GameCharacter.java
@@ -19301,6 +19301,14 @@ public abstract class GameCharacter implements XMLSaving {
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


Stadler76:
\~closing\~ here to make a PR that singles out commit 3536240582dde76b988fdfe5a2d58c359212d440

Issue #1224, opened 2019-10-19
Thin walls: sex in an apartment is public


AceXP:
**Describe the bug**
A companion moved out and into a Dominion tile. When starting sex there, it's marked as public sex and there is a line about spectators and their reaction to the performance.

This only happens right after the move (i.e. when first visiting the allocated tile). When you are visiting again everything works as expected.

**Expected behavior**
Sex in a companion's apartment should always be private.

**To Reproduce**

1. Have a companion move out to their own apartment.
2. Accompany them there (visit apartment option).
3. Start sex.

**Additional information:**
 - Game version [e.g. 0.3.5.1]
 - OS: Windows 10
 - Using .jar, Java: 1.8.0_202

Little-119:
It's Not A Bug, It's A Feature™

AceXP:
PR Set apartment to true when occupant moves out #1251 will fix this minor issue

Innoxia:
Fixed by merge of #1251

PR #1251, opened 2019-12-29
Set apartment to true when occupant moves out

AceXP:
Fix the minor issue with sex in an apartment being public when the occupant has just moved in

Set the boolean IsApartment to true in the dialogue when the occupant moves to the apartment

No

Yes, version 0.3.5.8


AceXP:
This will fix the minor issue Thin walls: sex in an apartment is public #1224

Innoxia:
Thank you!

PR #1253, opened 2019-12-29
Forced tf should fix leg config

Stadler76:
Forced TF didn't change the legs, if the leg configuration was incompatible

- The leg TF chunk of the forced TF checks, if the targeted LegConfiguration is compatible with the current one and sets the secondary TF modifier to set it to the default, if needed.
- Added `getDefaultLegConfiguration` to AbstractLegType.java: Returned the first entry in `allowedLegConfigurations` as the default.
- Added `getTFModifier` to the `LegConfiguration`s.

No

Version 0.3.5.8

Stadler#3007


Stadler76:
\~closing\~ because I can't change it to a draft and the WIP-prefix does nothing ...

PR #1260, opened 2020-01-01
CoverableArea mouth can never be discovered, blocking knowledge of NPC's oral skills

AceXP:

Fix the issue where an NPC's oral skill can never be discovered

CoverableArea mouth has saveDiscoveredArea set to false. 
There is a check in setAreaKnownByCharacter to see if it is true, and since it never is, 
this blocks gaining knowledge of this area by the MC

To fix this, the value has been set to true.

No

Yes, version 0.3.5.8


Innoxia:
Thank you for this!

PR #1263, opened 2020-01-02
Fix DayPeriodCalculation to prevent eternal night.

AceXP:

Fix the issue where the DayPeriod (daytime, twilight, nighttime) is not calculated correctly resulting in eternal night.

Use only time when determining DayPeriod, since the date can be (is always?) a day off. If the sunrise is calculated as time xx:xx tomorrow (as frequently happens), the current Dayperiod is always Night. The actual date used in the calculation (be it today, tomorrow or yesterday) doesn't matter all that much. It's comparing the current time with sunrise and sunset that's important.

No

Yes, version 0.3.5.8

@Innoxia: In public static LocalDateTime julianDayToDate(double julianDate) there was a plusDays(1) added in return LocalDateTime.of(year, month, day, hours, minutes, seconds).plusDays(1). Removing this again may also solve the problem, but I think it was put there for a reason?

### DIFF
--- a/src/com/lilithsthrone/controller/MainController.java
+++ b/src/com/lilithsthrone/controller/MainController.java
@@ -1544,8 +1544,8 @@ public class MainController implements Initializable {
 			TooltipInformationEventListener el2 = new TooltipInformationEventListener().setInformation(
 					(Main.game.isDayTime()?"Day-time":"Night-time"),
 					"Current time: "+Units.time(Main.game.getDateNow())+" (<span style='color:"+dp.getColour().toWebHexString()+";'>"+Util.capitaliseSentence(dp.getName())+"</span>)<br/>"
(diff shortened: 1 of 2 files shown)


Innoxia:
I did a temporary fix for this in the utils.time.DateAndTime.getTimeOfSolarElevationChange() method, so it should be working now (although with the possibility of the sunrise/set time being one day out, which isn't really an issue anyone will pick up on)...

I'll potentially revisit it at some point to see if I can get it fixed correctly.

AceXP:
The fix I did was to just look at the time portion of the day, that should always work, even if it's a day off. Also, the sunrise/sunset is shown in the calendar, and there it would be odd if the wrong day is shown. So time-only there too.

AceXP:
Never mind all that. You're patching the current date back into the calculated date, which has the same effect as my solution. So no need to change.

PR #1274, opened 2020-01-05
Changed text in prologue for MC as a trap.

AceXP:
Fix the issue where the NPC in the prologue knows about the MC's penis when this shouldn't be the case.

Add textual changes based on isFeminine, isPenisBulgeVisible and isTesticleBulgeVisible

No

Yes, version 0.3.5.8

There's a surprising number of possible combinations to try out!


Innoxia:
Thank you for this!

Issue #1275, opened 2020-01-06
Restocking of items sold by Vicky produces wrong prices.


zR1OQicz:
**Describe the bug**
Restocking of items sold by Vicky (Arcane Arts) produces wrong prices after buying all items of a single type.

**Expected behavior**
Either the prices of restocked items are correct or Vicky does restock those only at midnight like every other shop.

**To Reproduce**
1. Go to Vicky's shop.
2. Click on any item. See that the displayed prices are consistent.
![before buy](https://user-images.githubusercontent.com/58343091/71804459-c9937080-3063-11ea-887e-422c56c5d1bf.png)

3. Buy all items of that type by clicking 'Buy All'. See that the item is removed from the shop inventory after buying.
4. Leave and immediately reenter the shop. See that a single item has been restocked.
5. Click on the item. See that the price for one item is correct, but the price for 'Buy All' is 0.
![after buy](https://user-images.githubusercontent.com/58343091/71804474-d6b05f80-3063-11ea-8221-accb3e013a4d.png)

**Attach files**
error.log:
```
Version: 0.3.5.8
Java: 13.0.1
Warning: Nashorn engine is planned to be removed from a future JDK release
```

**Additional information:**
- Game version v0.3.5.8
- OS: SuSE Linux 64-bit
- self-compiled from github source, commit: 3a802445569eacd435d195cae8ad001e28b12c89
- java --version
  openjdk 13.0.1 2019-10-15
  OpenJDK Runtime Environment (build 13.0.1+9-suse-1.2-x8664)
  OpenJDK 64-Bit Server VM (build 13.0.1+9-suse-1.2-x8664, mixed mode)

Innoxia:
Fixed in merge of #1276

PR #1276, opened 2020-01-06
Fix empty stacks in Vicky's shop

zR1OQicz:

fix empty stacks/incorrect prices in Vicky's shop: #1275

I have Vicky remove all sold items from the ...ForSale lists when all items are sold.

no

yes, v0.3.5.8

none


Innoxia:
Thank you for this fix!

PR #1277, opened 2020-01-09
Avoid assumption of slime for debug body transform, unless already no…

kinsey-nsfw:

Avoid assumption of slime for debug body transform, unless already noted in comments as "slime/debug" or similar.

Changed "isDemonTFMenu()" in dialogue/utils/BodyChanging.java to "isDemonTFMenu() || debugMenu" in appropriate places.

No.

0.3.5.4

Kinsey/kinsey.


Innoxia:
I think that implementing the change as you've done it would inadvertently limit the amount of races available for self-TF in the debug menu. I've made some changes in the code on my end to address the issue of the debug menu treating the player as though they're a slime, so it's all sorted now. Thank you for bringing it my attention with your PR!

PR #1280, opened 2020-01-13
Merge with current innoxia dev version

kinsey-nsfw:
Will have to work out conflicts when I'm feeling up to it... sigh.


kinsey-nsfw:
Whoops! Wrong way around... will work on when I'm a bit more with it.

PR #1281, opened 2020-01-16
Add .gitattributes to normalize line endings

Stadler76:
Add a `.gitattributes` file so that line-endings are normalized.

Add `.gitattributes` and applied `.gitignore` changes suggested by `@bicobus#0479`

No.

No. It should affect only new and changed files. And gameplay is not affected by it.

`@Stadler#3007`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,28 @@
-* text=auto
+# Java sources
+*.java          text diff=java
+*.gradle        text diff=java
+*.gradle.kts    text diff=java
+
+# These files are text and should be normalized (Convert crlf => lf)
+*.css           text diff=css
+*.df            text
+*.htm           text diff=html
+*.html          text diff=html
+*.js            text
+*.jsp           text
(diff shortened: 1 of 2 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto
```
bicobus: If you set text=auto for all files, you need to ensure that binary files (like images) are excluded. Otherwise, git might detect a .jpg as a text file and try to normalize the line endings. It would be better to limit the normalization to .java files only.

See https://www.git-scm.com/docs/gitattributes#_effects

kinsey-nsfw:
I'm currently having quite a bit of headaches with line endings myself - setting renormalize = true is not sufficient to fix the problem of PC vs Mac/Unix (the latter being what I use). I do have a git attributes file in my .config/git directory:

*       text=auto
*.ps    binary
*.pdf binary
*.jar binary
*.class binary
*.png binary
*.jpg binary
*.gif binary
*.ttf binary
*.svg delta diff merge=text
*.java text diff=java merge=text
*.xml text diff merge=text
*.yml text diff merge=text
*.css text diff=css merge=text
*.md text diff merge=text
*.txt text diff merge=text

PR #1286, opened 2020-01-24
FaceTypeTag implementation

Stadler76:
Add tags to bodyparts (here to the FaceType) to avoid unneeded complexity, when certain features can be answered with simple yes/no questions. Like:
- Does the creature have a muzzle?
- Is it natural for the creature to have no hair on its head?

Added a few `FaceTypeTag`s and implemented them to show, how they could be used.
Current usages:
- Feral Bite excluded some `FaceType`s before that. Now its enabled, when the creature has a muzzle, fangs, shark teeth or a beak. Gameplay wise nothing has changed here.
- If its natural for that creature to have no hair on the head the appearance doesn't mention it and instead the body covering text is printed:
  Before: before
  After: after

Currently I have added a simple enum with no extra fields. But this could be improved later.
I've added the field `private List<FaceTypeTag> tags;` to the `FaceType` enum and added the proper tags to the face types, where they are applicable (Did some recherche for horses, hance the comment there).

No.

Yes. Brief test with 0.3.6.1 and working as intended.

`@Stadler#3007`


Stadler76:
I've improved the PR to show up 'Hair: None' in the race tooltip and 'You now have **no hair**!' in the TF blurb, when having no hair is natural.

The following methods were added for this:
- In `src/com/lilithsthrone/game/character/body/Face.java`:
  ```java
  public boolean isBaldnessNatural() { 
  	return this.getTypeTags().contains(FaceTypeTag.NATURAL_BALDNESS); 
  }
  ```
- In `src/com/lilithsthrone/game/character/GameCharacter.java`:
  ```java
  public boolean isFaceBaldnessNatural() { 
  	return body.getFace().isBaldnessNatural(); 
  } 
  ```

Example screenshot showing the changes:
Example screenshot showing the changes

src/com/lilithsthrone/game/character/body/tags/FaceTypeTag.java
```diff
@@ -0,0 +1,21 @@
...
+	MUZZLE,
+	/** Fangs instead of normal teeth */
+	FANGS,
+	/** Beaked faces */
+	BEAK,
+	/** Pointy, trianguler shark teeth (Currently NYI, but planned for the (far?) future) */
+	SHARK_TEETH,
+	/** Being bald aka hairLength == 0 is not unnatural for that faceType */
+	NATURAL_BALDNESS,
+}
```
Stadler76: I could add some parser commands to be used in conditionals for that.
Suggestions:
- `npc.hasMuzzle`
- `npc.hasFangs`
- `npc.hasBeak`
- `npc.hasSharkTeeth`
- `npc.isFaceBaldnessNatural`, `npc.isBaldnessNatural`, `npc.isFaceNaturallyBald` and `npc.isNaturallyBald`

Innoxia: I'll get those added after merging it; good idea! ^^

Innoxia: (I've added them as methods in the GameCharacter class, which can be called in parsing conditionals.)

Issue #1295, opened 2020-02-08
Inflation Mechanic Updated


ThyHolyDevil:
Before I get into updated mechanic, I'm pretty sure the way I'm posting this is inefficient. I'm unfamiliar with using github, better ways are welcome.

**File Changed:**
src\com\lilithsthrone\game\character\effects\StatusEffect.java

**>>Change 1:**
Lines = 5288, 5289
```
			return cumAmount >= CumProduction.SEVEN_MONSTROUS.getMinimumValue()
					&& cumAmount < CumProduction.SEVEN_MONSTROUS.getMedianValue()
```
Replaced With:
```
			return cumAmount >= Math.pow((Math.cbrt(3785) / 182.88) * target.getHeightValue(), 3)
					&& cumAmount < Math.pow((Math.cbrt(7570) / 182.88) * target.getHeightValue(), 3)
```
**>>Change 2:**
Lines = 5328, 5329
```
			return cumAmount >= CumProduction.SEVEN_MONSTROUS.getMedianValue()
					&& cumAmount < CumProduction.SEVEN_MONSTROUS.getMaximumValue()
```
Replaced With:
```
			return cumAmount >= Math.pow((Math.cbrt(7570) / 182.88) * target.getHeightValue(), 3)
					&& cumAmount < Math.pow((Math.cbrt(11356) / 182.88) * target.getHeightValue(), 3)
```
**>>Change 3:**
Line = 5368
```
			return cumAmount >= CumProduction.SEVEN_MONSTROUS.getMaximumValue()
```
Replaced With:
```
			return cumAmount >= Math.pow((Math.cbrt(11356) / 182.88) * target.getHeightValue(), 3)
```
**>>Change 4:**
Lines = 5407, 5408
```
			return cumAmount >= CumProduction.SEVEN_MONSTROUS.getMinimumValue()
					&& cumAmount < CumProduction.SEVEN_MONSTROUS.getMedianValue()
```
Replaced With:
```
			return cumAmount >= Math.pow((Math.cbrt(1250) / 182.88) * target.getHeightValue(), 3)
					&& cumAmount < Math.pow((Math.cbrt(2500) / 182.88) * target.getHeightValue(), 3)
```
**>>Change 5:**
Lines = 5447, 5448
```
			return cumAmount >= CumProduction.SEVEN_MONSTROUS.getMedianValue()
					&& cumAmount < CumProduction.SEVEN_MONSTROUS.getMaximumValue()
```
[...]

PR #1296, opened 2020-02-08
StatusEffect.java - Inflation Mechanic Updated

ThyHolyDevil:
**Why I Do:**
Was making a personalized, modded version of game for self-use, reading through code.
Saw the usage of a static value, decided to fix.

**What It Do:**
Made the fluid volume properly scale with character height.

**How It Do:**
1)Cube-Rooting as a quick way to turn a fluid volume into a 3D scalable number
2)dividing by the height I had based fluid volume around. (6ft, 182.88cm)
3)Multiplying by affected character's height, to create scalability
4)power of 3 to De-Cube-Root, converting from 3D scalable number back into fluid volume.
(Feel free to change Fluid Volume in cbrt() or change the reference height.)

**It Do Comparisons:**
VARS:
A= 3785; (Fluid for 6ft / 182.88cm, 1 US Gallon in mL)
B= ##; (New Height to calc For)

**Calc for smallest extreme:** (2ft, B=61)
>Option 1: (basic Divide & Mult)
(A / 182.88) * B = 1,262 ml (0.3333 US Gal, way too much)
>Option 2: (Square Root)
((sqrt(A)/182.88)*B)^2 = 421 ml (0.1112 US Gal, still too much)
>Option 3: (Cube Root)
((cbrt(A)/182.88)*B)^3 = 140 ml (0.037 US Gal / 0.59 US Cup)

**Calc for avg height:** (6ft, B=182.88) [just for showing, otherwise useless]
>Option 1: (basic Divide & Mult)
(A / 182.88) * B = 3,785 ml (1 US Gal)
>Option 2: (Square Root)
((sqrt(A)/182.88)*B)^2 = 3,785 ml (1 US Gal)
>Option 3: (Cube Root)
((cbrt(A)/182.88)*B)^3 = 3,785 ml (1 US Gal)

**Calc for largestest extreme:** (12ft, B=366)
>Option 1: (basic Divide & Mult)
(A / 182.88) * B = 7,575 ml (2 US Gal, not enough)
>Option 2: (Square Root)
((sqrt(A)/182.88)*B)^2 = 15,160 ml (4 US Gal, enough? not really?)
>Option 3: (Cube Root)
((cbrt(A)/182.88)*B)^3 = 30,340 ml (8 US Gal, enough.)

**What It Could Do:** (Untested)
IF
target.getHeightValue()
INTO
(target.getHeightValue() * ((Bodysize + 100) / 150))
Results:
BodySize= 0 - would mult height by 0.6666...
BodySize= 50 - would mult height by 1, no change
BodySize= 100 - Would mult height by 1.3333...

**For Nipple&Breast inflation**
I decided to do a quicker method than what I was originally thinking of. I only did rough tuning. Feel free to adjust the number that is in 'cbrt()' and the divisor for breast size(by 0.1 increments probably)

Here's the math I used in my calculator: I'm too lazy to make more readable right now.
F=10; //Fluid base
CS=16; //Cup Size
DV=3; // Divisor
SN=(3#F); //Cube Rooting Fluid base
SF=SN * ((CS/DV)+1); // Smash the numbers together
SF^3 // De-Cube-Root to get result

**End Notes:**
This is just a quick fix to a lack of scaling. You could turn my math into function elsewhere and simplify process.

The Cube-Root process can be used elsewhere, You could overhaul Girth, Capacity, ETC using this math process.

Updating Static values.
Added Math to Inflation Mechanics
No?
The suggested changes were lightly tested, code in description is probably not tested.
HolyDevil#6986
Probably will do later.

### DIFF
--- a/src/com/lilithsthrone/game/character/effects/StatusEffect.java
+++ b/src/com/lilithsthrone/game/character/effects/StatusEffect.java
@@ -5263,8 +5263,8 @@ public enum StatusEffect {
 		@Override
 		public boolean isConditionsMet(GameCharacter target) {
 			float cumAmount = target.getTotalFluidInArea(SexAreaOrifice.ANUS) + target.getTotalFluidInArea(SexAreaOrifice.MOUTH) + target.getTotalFluidInArea(SexAreaOrifice.VAGINA);
-			return cumAmount >= CumProduction.SEVEN_MONSTROUS.getMinimumValue()
-					&& cumAmount < CumProduction.SEVEN_MONSTROUS.getMedianValue()
+			return cumAmount >= Math.pow((Math.cbrt(3785) / 182.88) * target.getHeightValue(), 3)
+					&& cumAmount < Math.pow((Math.cbrt(7570) / 182.88) * target.getHeightValue(), 3)
 					&& Main.getProperties().hasValue(PropertyValue.inflationContent);
 		}
 		
@@ -5303,8 +5303,8 @@ public enum StatusEffect {
 		@Override
 		public boolean isConditionsMet(GameCharacter target) {
 			float cumAmount = target.getTotalFluidInArea(SexAreaOrifice.ANUS) + target.getTotalFluidInArea(SexAreaOrifice.MOUTH) + target.getTotalFluidInArea(SexAreaOrifice.VAGINA);
-			return cumAmount >= CumProduction.SEVEN_MONSTROUS.getMedianValue()
-					&& cumAmount < CumProduction.SEVEN_MONSTROUS.getMaximumValue()
+			return cumAmount >= Math.pow((Math.cbrt(7570) / 182.88) * target.getHeightValue(), 3)
+					&& cumAmount < Math.pow((Math.cbrt(11356) / 182.88) * target.getHeightValue(), 3)
 					&& Main.getProperties().hasValue(PropertyValue.inflationContent);
 		}
 		
@@ -5343,7 +5343,7 @@ public enum StatusEffect {
 		@Override
 		public boolean isConditionsMet(GameCharacter target) {
 			float cumAmount = target.getTotalFluidInArea(SexAreaOrifice.ANUS) + target.getTotalFluidInArea(SexAreaOrifice.MOUTH) + target.getTotalFluidInArea(SexAreaOrifice.VAGINA);
-			return cumAmount >= CumProduction.SEVEN_MONSTROUS.getMaximumValue()
+			return cumAmount >= Math.pow((Math.cbrt(11356) / 182.88) * target.getHeightValue(), 3)
 					&& Main.getProperties().hasValue(PropertyValue.inflationContent);
 		}
 		
@@ -5382,8 +5382,8 @@ public enum StatusEffect {
 		@Override
 		public boolean isConditionsMet(GameCharacter target) {
 			float cumAmount = target.getTotalFluidInArea(SexAreaOrifice.NIPPLE);
-			return cumAmount >= CumProduction.SEVEN_MONSTROUS.getMinimumValue()
-					&& cumAmount < CumProduction.SEVEN_MONSTROUS.getMedianValue()
+			return cumAmount >= Math.pow((Math.cbrt(10) * (target.getBreastSize().getMeasurement() / 3) + 1), 3)
+					&& cumAmount < Math.pow((Math.cbrt(20) * (target.getBreastSize().getMeasurement() / 3) + 1), 3)
 					&& Main.getProperties().hasValue(PropertyValue.inflationContent);
 		}
 		
@@ -5422,8 +5422,8 @@ public enum StatusEffect {
 		@Override
 		public boolean isConditionsMet(GameCharacter target) {
 			float cumAmount = target.getTotalFluidInArea(SexAreaOrifice.NIPPLE);
-			return cumAmount >= CumProduction.SEVEN_MONSTROUS.getMedianValue()
-					&& cumAmount < CumProduction.SEVEN_MONSTROUS.getMaximumValue()
+			return cumAmount >= Math.pow((Math.cbrt(20) * (target.getBreastSize().getMeasurement() / 3) + 1), 3)
+					&& cumAmount < Math.pow((Math.cbrt(30) * (target.getBreastSize().getMeasurement() / 3) + 1), 3)
 					&& Main.getProperties().hasValue(PropertyValue.inflationContent);
 		}
 		
@@ -5462,7 +5462,7 @@ public enum StatusEffect {
 		@Override
 		public boolean isConditionsMet(GameCharacter target) {
 			float cumAmount = target.getTotalFluidInArea(SexAreaOrifice.NIPPLE);
-			return cumAmount >= CumProduction.SEVEN_MONSTROUS.getMaximumValue()
+			return cumAmount >= Math.pow((Math.cbrt(30) * (target.getBreastSize().getMeasurement() / 3) + 1), 3)
 					&& Main.getProperties().hasValue(PropertyValue.inflationContent);
 		}
 		


ThyHolyDevil:
Just mentioning, I have no clue why I removed the Height Scaling from Breast Inflation... It technically still makes sense.

TadUnlikely:
Associated with issue #1295

ThyHolyDevil:
@Rfpnj
### Clarify please?
I'm relatively new to coding, and am entirely self-taught. On top of that, I'm trying to learn python, not java. So, I'll be running on assumptions. If I say something stupid, please let me know.

**_->Were these changes traced back to see how the code worked prior to changes?_**
**Before:**
>CumProduction.SEVEN_MONSTROUS.getMinimumValue()
>CumProduction.SEVEN_MONSTROUS.getMedianValue()
>CumProduction.SEVEN_MONSTROUS.getMaximumValue()

Those were used to determine how much fluid is needed for each tier of inflation. These methods called a static number. I don't like the idea of calling a value from something else that is not 100% relevant. The amount one needs to be inflated with should not be bound by the amount one can put out.

I replace those static value calls with math that allows for the amount needed to change depending on the height of the afflicted., and for breasts, I even added math to scale with the size of the breasts. That is all I've done _for now_.

**_->As it looks the changes would orphan methods that should then be removed._**
I'm not entirely sure, but if you mean unused code, I think they still get used by NPC creation data.

I also wanna clarify the code in the pull request has been tested on latest DEV build., if the code is in the description, it's just me spilling ideas around, and is untested.

src/com/lilithsthrone/game/character/effects/StatusEffect.java
```diff
@@ -5263,8 +5263,8 @@ public String extraRemovalEffects(GameCharacter target) {
 		@Override
 		public boolean isConditionsMet(GameCharacter target) {
 			float cumAmount = target.getTotalFluidInArea(SexAreaOrifice.ANUS) + target.getTotalFluidInArea(SexAreaOrifice.MOUTH) + target.getTotalFluidInArea(SexAreaOrifice.VAGINA);
-			return cumAmount >= CumProduction.SEVEN_MONSTROUS.getMinimumValue()
-					&& cumAmount < CumProduction.SEVEN_MONSTROUS.getMedianValue()
+			return cumAmount >= Math.pow((Math.cbrt(3785) / 182.88) * target.getHeightValue(), 3)
```
Eagle0600: This calculation is almost equal to `Math.pow(target.getHeightValue(), 3) / 1616` (~0.002% difference) and is a much simpler calculation.

src/com/lilithsthrone/game/character/effects/StatusEffect.java
```diff
@@ -5303,8 +5303,8 @@ public String extraRemovalEffects(GameCharacter target) {
 		@Override
 		public boolean isConditionsMet(GameCharacter target) {
 			float cumAmount = target.getTotalFluidInArea(SexAreaOrifice.ANUS) + target.getTotalFluidInArea(SexAreaOrifice.MOUTH) + target.getTotalFluidInArea(SexAreaOrifice.VAGINA);
-			return cumAmount >= CumProduction.SEVEN_MONSTROUS.getMedianValue()
-					&& cumAmount < CumProduction.SEVEN_MONSTROUS.getMaximumValue()
+			return cumAmount >= Math.pow((Math.cbrt(7570) / 182.88) * target.getHeightValue(), 3)
```
Eagle0600: This calculation is almost equal to `Math.pow(target.getHeightValue(), 3) / 808` (~0.002% difference) and is a much simpler calculation.

src/com/lilithsthrone/game/character/effects/StatusEffect.java
```diff
@@ -5343,7 +5343,7 @@ public String extraRemovalEffects(GameCharacter target) {
 		@Override
 		public boolean isConditionsMet(GameCharacter target) {
 			float cumAmount = target.getTotalFluidInArea(SexAreaOrifice.ANUS) + target.getTotalFluidInArea(SexAreaOrifice.MOUTH) + target.getTotalFluidInArea(SexAreaOrifice.VAGINA);
-			return cumAmount >= CumProduction.SEVEN_MONSTROUS.getMaximumValue()
+			return cumAmount >= Math.pow((Math.cbrt(11356) / 182.88) * target.getHeightValue(), 3)
```
Eagle0600: This calculation is almost equal to `Math.pow(target.getHeightValue(), 3) / 539` (~0.07% difference) and is a much simpler calculation.

src/com/lilithsthrone/game/character/effects/StatusEffect.java
```diff
@@ -5382,8 +5382,8 @@ public String extraRemovalEffects(GameCharacter target) {
 		@Override
 		public boolean isConditionsMet(GameCharacter target) {
 			float cumAmount = target.getTotalFluidInArea(SexAreaOrifice.NIPPLE);
-			return cumAmount >= CumProduction.SEVEN_MONSTROUS.getMinimumValue()
-					&& cumAmount < CumProduction.SEVEN_MONSTROUS.getMedianValue()
+			return cumAmount >= Math.pow(((Math.cbrt(10)/182.88)*target.getHeightValue())*((target.getBreastSize().getMeasurement()/3)+1),3)
```
Eagle0600: This calculation is almost equal to `Math.pow(target.getHeightValue()*((target.getBreastSize().getMeasurement()/3)+1), 3) / 611644` (~0.00002% difference) and is a much simpler calculation.

src/com/lilithsthrone/game/character/effects/StatusEffect.java
```diff
@@ -5422,8 +5422,8 @@ public String extraRemovalEffects(GameCharacter target) {
 		@Override
 		public boolean isConditionsMet(GameCharacter target) {
 			float cumAmount = target.getTotalFluidInArea(SexAreaOrifice.NIPPLE);
-			return cumAmount >= CumProduction.SEVEN_MONSTROUS.getMedianValue()
-					&& cumAmount < CumProduction.SEVEN_MONSTROUS.getMaximumValue()
+			return cumAmount >= Math.pow(((Math.cbrt(20)/182.88)*target.getHeightValue())*((target.getBreastSize().getMeasurement()/3)+1),3)
```
Eagle0600: This calculation is almost equal to `Math.pow(target.getHeightValue()*((target.getBreastSize().getMeasurement()/3)+1), 3) / 305822` (~0.00002% difference) and is a much simpler calculation.

src/com/lilithsthrone/game/character/effects/StatusEffect.java
```diff
@@ -5462,7 +5462,7 @@ public String extraRemovalEffects(GameCharacter target) {
 		@Override
 		public boolean isConditionsMet(GameCharacter target) {
 			float cumAmount = target.getTotalFluidInArea(SexAreaOrifice.NIPPLE);
-			return cumAmount >= CumProduction.SEVEN_MONSTROUS.getMaximumValue()
+			return cumAmount >= Math.pow(((Math.cbrt(30)/182.88)*target.getHeightValue())*((target.getBreastSize().getMeasurement()/3)+1),3)
```
Eagle0600: This calculation is almost equal to `Math.pow(target.getHeightValue()*((target.getBreastSize().getMeasurement()/3)+1), 3) / 203881` (~-0.00014% difference) and is a much simpler calculation.

ThyHolyDevil:
@Eagle0600 Sorry but frankly, your math does not scale similarly, and is not worth it computationally. 

You're making the code harder to understand its purpose, while reducing its scaling accuracy, for practically no performance gain.

Also, I tried testing your version of the breast inflation mechanics, and was getting unreasonably low numbers. You may have made an error in your syntax.

As far as I'm concerned, A computer has no issues calculating my math, and can even be converted into a function elsewhere for simplicity, and reusability. To reuse your math, each instance would require your divider to be recalculated, which frankly is a waste of time for again, practically no performance gain.

------

Don't try to fix what is not broken. As far as I'm concerned the only thing broken about my offering is that the math is copy-pasted for each instance, while it could be turned into a function elsewhere, simplifying the code further.

Eagle0600:
> @Eagle0600 Sorry but frankly, your math does not scale similarly, and is not worth it computationally.

Algebraicly, that's wrong. It should scale identically, because a³ × b³ = (a × b)³, and the only cube-root in there is a constant. The only difference should be a constant difference due to the constants I used being rounded, and that difference is insignificant.

ThyHolyDevil:
@Eagle0600 "math does not scale similarly" - "reducing its scaling accuracy" I'm not saying it is completely off. I know how the math works. If anything I probably shouldn't be using the term scaling in those lines, as that's probably a source for confusion.

Here's the big issue. With my math you can read it and get an idea of what it is trying to do, and swap variables with ease for different purposes. Aka if converted into a function, It would be practically child's play to use in different locations.

With your math, You consolidated all variables into one number that is completely obscure to its purpose _for most people_. And even for people who understand it, Its far from plugging in some values and being done with it.

All in all I have no interest at all in using your math, as It just makes it harder to adjust it should I need to, it is unsuitable for a function, and just frankly not worth it.

CognitiveMist:
I am extremely hesitant to add "magic values". That is, bare numbers with no label or explanation. In my professional experience, magic values make code much, much harder to maintain.

In the existing code, there are no visible numbers, and instead named values are used that describe exactly what is going on. For example:

    cumAmount < CumProduction.SEVEN_MONSTROUS.getMedianValue()

This compares the cum amount to the median value for the monstruous-level of cum production. The code is self-documenting, and little to no effort is required to understand what's going on. By comparison:

    cumAmount < Math.pow((Math.cbrt(7570) / 182.88) * target.getHeightValue(), 3)

This compares the cum amount to some strange math based on the height, but it's hard to understand what exactly it is doing without sitting down to decipher it.

This would be improved by a massive amount if you could refactor this to be something along the lines of:

    cumAmount < Math.pow((Math.cbrt(CumProduction.SEVEN_MONSTROUS.getMedianVolume()) / CumProduction.SEVEN_MONSTROUS.getMedianHeight()) * target.getHeightValue(), 3)

Or even better:

    cumAmount < Math.pow(CumProduction.SEVEN_MONSTROUS.getMedianHeightConversionFactor() * target.getHeightValue(), 3)

Or better yet:

    cumAmount < CumProduction.SEVEN_MONSTROUS.getMedianConversionFactor() * target.getBodyVolume()

Note in this last example, the calculation is just a single multiplication, and the function getBodyVolume() can easily be made to take the target's build into account as well. (There may already be a similar function in the game, I haven't looked. Also, a function getBodyMass() would be equally effective here.)

ThyHolyDevil:
@CognitiveMist
I actually agree completely. I submitted this more as an example of a way to achieve inflation scaling that makes sense.

End goal would be to turn all the math into a function, And store the variables with function so make a file, 'inflation.java' lets say, and store variables for each tier of inflation, along with the math and calls to use that math so in the end the usage would be something like:
cumamount > Inflation.BELLY.getSlightBloat()
cumamount > Inflation.BELLY.getModerateBloat()
cumamount > Inflation.BELLY.getExcessiveBloat()
-

PR #1298, opened 2020-02-10
Fix Amber's level drain

triples941:
Fixes inability to orgasm when having sex with Amber with no neck clothing, by adding a null check.
Fix checked on v0.3.6.8
Discord Handle: TripleS941#0897


Innoxia:
Thank you!

PR #1305, opened 2020-02-27
Removing Endline Spaces

klorpa:
Removes unnecessary endline spaces.


AceXP:
Did you actually build this? Because it looks like you are deleting files that are needed in the game?

CognitiveMist:
Usually whitespace-only changes are not very useful, since they dirty the git history without improving the behavior of the program.

If you want to fix the whitespace in and around a different change you are making, that is generally fine.

However, it also looks like a bunch of other stuff snuck in to this change, that probably was unintentional, based on your description. See src/com/lilithsthrone/game/sex/positions/slots/SexSlotDesk.java, for example.

Part of the problem may be that you've based this change off of `master`, with the intent of merging it into `dev`. In fact (at least in the world of LT), you should base your changes off `dev`, not `master`.

Innoxia:
Thank you for your work on this PR, but there are too many conflicts for it to be merged at this point (and I don't think the removal of endline spaces is a pressing issue in any case). Sorry for how long it took me to respond to this!

PR #1312, opened 2020-03-04
Enable changing your hair style in your own room

AceXP:
Add an option to change your hair style in your own room.

Add a sub-option to the bathroom. Shows a shortened list of hair styles (only hair styles that are deemed simple enough to change yourself). Otherwise same functionality as with Kate. 

No

Yes, tested in version 3.6.9

ACEXP


AceXP:
To be replaced by new version based on latest build.

Issue #1304, opened 2020-02-26
Slime main color not saved


AlleMazza:
**Description**
The main color of slime characters (the one regarding torso/arms/legs and any other part that can't have a different color) is not saved. Each time a slime is loaded, it randomly changes, even if they are wearing an anti-transformation item.
Adding it manually to the save file seems to work. I tried adding
`<bodyCovering c1="COVERING_PURPLE_DARK" c2="COVERING_PURPLE_DARK" discovered="true" modifier="GOOEY" pattern="NONE" type="SLIME"/>`
to the list of body coverings on a slime slave, and now the color is static. Even saving the game in a new save file keeps the edit. Also, after the edit, if I change the slime's color using the transformation menu, the new color is saved properly.

**To Reproduce**
Go into Submission, explore until finding a slime, fight it without removing it, save, load, and the color changes.

**Attach files**
Here's a save file with two slime slaves, one save-edited (Francy), and one left as the game generated it (Taylor). Every time you load this save, Taylor's color changes, Francy's doesn't.
[New Save.zip](https://github.com/Innoxia/liliths-throne-public/files/4255293/New.Save.zip)

**Additional information:**
 - Game version 0.3.6.8
 - OS: Windows 7
 - Using .jar or .exe: .exe

AceXP:
This has now been fixed in version 0.3.9.1, please check it and close this issue

PR #1313, opened 2020-03-04
Add SLIME to discovered coverings (also when loading)

AceXP:
Fix issue #1304: The main color of slime characters is not saved.

Add SLIME as a body covering part, so it will also be added to the save file. And on loading back the save, also update the list with coverings.

No

Yes, version 0.3.6.9

ACEXP


Innoxia:
Thank you for this PR! I've implemented it in my local code (as it was quicker than resolving the conflicts and pulling), and I've credited you in the patch notes for this fix. ^^

PR #1320, opened 2020-03-18
Reduce length of offspring list.

karappoki:
Shortening the offspring list under pregnancy stats by removing the children which were removed from the game after attacking them.

PlayerCharacter.java:
   Added new variable to store the number of removed children
   Added getter and increment function for new variable
   Updated save and load function to include the new variable
PhoneDialogue.java:
   Updated calculation method for percentage of already met children
Game.java:
   Added a bit of code to the function for removing NPCs, that removes the NPC, if related to the player, from NPCmap and increases the new variable by 1.

No.

I tested it briefly with 0.3.6.9 Alpha.


Innoxia:
This issue seems to have been fixed at some point after you submitted this PR. Thank you for your work regardless, and I'm sorry that it took me so long to respond to this!

PR #1333, opened 2020-04-29
Fluid Control

The7thWolf:

Adding my mod "Fluid Control" to the base game, as suggested by a few members of the community.

This PR adds buttons in the milking room that let you choose the amount of liquid that is pumped out of the tanks. It is capped at 5L, regardless of any installed room upgrades.
fluid-control

No

Tested with versions from 0.3.2.2 all the way to the current 0.3.7.3.

TheSeventhWolf#8843

---
I'd like to warn you that I am not a Java dev and basically just reverse engineered the whole thing. I don't expect any bugs caused by this PR though.

### DIFF
--- a/src/com/lilithsthrone/controller/MainControllerInitMethod.java
+++ b/src/com/lilithsthrone/controller/MainControllerInitMethod.java
@@ -1405,6 +1405,99 @@ public class MainControllerInitMethod {
 			if(Main.game.getPlayer().getLocationPlace().getPlaceUpgrades().contains(PlaceUpgrade.LILAYA_MILKING_ROOM)) {
 				MilkingRoom room = Main.game.getOccupancyUtil().getMilkingRoom(Main.game.getPlayerCell().getType(), Main.game.getPlayerCell().getLocation());
 
+				// Fluid amount button listeners:
+				id = "AMOUNT_MINUS_ONE";
+				if (((EventTarget) MainController.document.getElementById(id)) != null) {
+					((EventTarget) MainController.document.getElementById(id)).addEventListener("click", e -> {
+						if (MilkingRoom.INGESTION_AMOUNT >= 6) {
+							MilkingRoom.INGESTION_AMOUNT = MilkingRoom.INGESTION_AMOUNT - 1;
+						}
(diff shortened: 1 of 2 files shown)


src/com/lilithsthrone/controller/MainControllerInitMethod.java
```diff
@@ -1405,6 +1405,99 @@ public void effects() {
 			if(Main.game.getPlayer().getLocationPlace().getPlaceUpgrades().contains(PlaceUpgrade.LILAYA_MILKING_ROOM)) {
 				MilkingRoom room = Main.game.getOccupancyUtil().getMilkingRoom(Main.game.getPlayerCell().getType(), Main.game.getPlayerCell().getLocation());
 
+				// Fluid amount button listeners:
+				id = "AMOUNT_MINUS_ONE";
+				if (((EventTarget) MainController.document.getElementById(id)) != null) {
+					((EventTarget) MainController.document.getElementById(id)).addEventListener("click", e -> {
+						if (MilkingRoom.INGESTION_AMOUNT >= 6) {
+							MilkingRoom.INGESTION_AMOUNT = MilkingRoom.INGESTION_AMOUNT - 1;
```
CognitiveMist, line 1413: Doesn't really matter, but you could shorten these lines like this:
```suggestion
							MilkingRoom.INGESTION_AMOUNT -= 1;
```
(and use `+=` for addition)

The7thWolf:
> 
> 
> According to the screenshot, the buttons seem to be:
> -1, -10, -100, -1000, +1, +10, +100, +1000
> 
> According to the code it looks like you've changed it to:
> -1000, -100, -10, -1, +1, +10, +100, +1000
> 
> Can you confirm that the screenshot is out of date? Because I do like the second ordering better.

Yes, you're right. I didn't even notice. The screenshot is out of date.
I changed the ordering as I didn't like the old ordering aswell.

src/com/lilithsthrone/controller/MainControllerInitMethod.java
The7thWolf: You're right. Done.

PR #1334, opened 2020-04-29
Fixed alternating sizes in forced TF.

Stadler76:

**Hair length, breast size, ass size and hip size had no tolerance so that forced TF alternated between too big and too small. Fixed that.**

**See above.**

**No.**

**Yes, 0.3.7.3 Alpha**

**Stadler#3007**


CognitiveMist (approved):

Innoxia:
Thank you for this!

Issue #1337, opened 2020-05-05
Equipping a unique NPC with a strapon results in a Nullpointer when sex ends


AceXP:
**Describe the bug**
Equipping a unique NPC (e.g. Bree/Brandi) with a strapon results in a Nullpointer exception when sex ends. This happens because for unique NPC's there is a check to put their equipment/clothing back in order. However, this check results in an error because the strapon is in a location where there was no equipment before.

**Expected behavior**
The game should continue without an error.

**To Reproduce**
1. Have a unique NPC as a companion. 
2. Start sex with this companion.
3. Equip a strapon on them.
4. When sex ends the game will not respond and there will be a Nullpointer exception.

**Attach files**
[error.log](https://github.com/Innoxia/liliths-throne-public/files/4580799/error.log)

**Additional information:**
 - Game version 0.3.7.3
 - OS: Windows 10
 - Using .jar 
- java version: 1.8.0_241

AceXP:
Fixed in version 0.3.9

PR #1338, opened 2020-05-05
Fix nullpointer with strapon

AceXP:
Fix a Nullpointer exception after equipping a unique NPC with a strapon. When sex ends, looping over all equipped clothing will throw a Nullpointer because there is no entry for the item in the penis slot in the original map. (Issue #1337).

Added a check to prevent a Nullpointer when equipped clothing is in a slot that wasn't equipped before. In this case, the new equipment is put back into the inventory.

No.

Yes, version 0.3.7.3

Acexp

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ target/
 
 # Ignore all generic character images
 res/images/characters/generic/
+/tail.txt
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,3 @@ target/
 
 # Ignore all generic character images
 res/images/characters/generic/
-/tail.txt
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -12,3 +12,4 @@ target/
 
 # Ignore all generic character images
 res/images/characters/generic/
+/tail.txt
```
Stadler76, line 15: I guess, this can be removed?

AceXP: Yes, that shouldn't be there. Removed it now.

AceXP:
Version 0.3.9 fixes this issue, so this PR can be closed

PR #1343, opened 2020-05-21
Fixed wrong pronoun parser var in potion text

Eliria:
Changed "npc" to "npc2" to reference the one taking the potion pronouns. Was saying 'you starts to feel' instead of he/she starts when you give others the tf potion.

Minor Textual / Parser ref fix.

Did not compile a new build for such a minor edit. Should be verifiable on reading.

Eliria


turkeyleg92:
Why is this not approved yet?...

Innoxia:
Thank you for this!

PR #1354, opened 2020-06-10
Fixed 'missing' harpy content errors, path referring to alleywayAttack

Eliria:
Some harpy-specific entries were throwing a missing content error because it was using a literal path pointing to _alleywayAttack.xml_ instead of the _getFileLocation_ function to access the harpy files.

Fixing harpy-zone dialog entries not being properly pulled from the harpy-specific dialog files.

- What you changed or added.
**A)** Made _harpyAttackerDialog_ use _getFileLocation()_ to uniformly obtain entries.
Some of the harpy-specific entries were trying to look in _alleywayAttacker.xml_ incorrectly. Many of the more generic ones also used a literal path for '_encounter/dom/alleywayAttack_' rather than their proper source determined by _getFileLocation()_.
**B)** Because of **A** some generic responses for harpy dialog entries during a storm started throwing missing dialog errors.  Method of fixing this was to duplicate the entries out of _harpyAttack.xml_ that were still valid during a storm into _harpyAttackStorm,xml_.
Downside is that this makes any entries that are identical between the alleywayattack/harpy/storm files annoying to edit instead of having a default base to refer to. Think that the getFileLocation() should be improved to handle any needed tweaking of the content file access, rather than using a mix-match in how the file path is given to the response code. Will continue poking at improving this.

No

It's been tested fairly thoroughly in 3.8.1 to cover all previous (or caused) harpy missing dialog errors. Looks good.

Eliria

### DIFF
--- a/res/txt/encounters/dominion/harpyAttackStorm.xml
+++ b/res/txt/encounters/dominion/harpyAttackStorm.xml
@@ -437,6 +437,892 @@
 	</p>
 	]]>
 	</htmlContent>
+<!-- Duplicated from harpyAttack ... -->
+
+	<!-- HARPY_PEACEFUL_ATTACK -->
+	
+	<htmlContent tag="HARPY_PEACEFUL_ATTACK"><![CDATA[
+	<p>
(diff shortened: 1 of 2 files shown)


Innoxia:
Thank you for this PR, but I've fixed it on my end by simply replacing the _dominion/alleywayAttack_ parts with _dominion/harpyAttack_. (The way that I've set these encounter dialogues nodes up is so that only the dialogue which has 'storm' variations need to use the getFileLocation() method.)

PR #1365, opened 2020-07-18
wrong feminity description

lure0xaos:
wrong feminity description "very masculine" -> "very feminine"


Innoxia:
I've got this fixed in the code on my end already, but thank you for the PR regardless!

PR #1368, opened 2020-07-23
String Match Enhancements

CognitiveMist:
### Changes
- Added unordered-word matching, so that RED_DARK can match DARK_RED.
- Added fixed prefixes for the unordered word matching, so that with
  a prefix of one, the match for ICE_CREAM would be ACE_DREAM, not
  CREME_FRAICHE. (The unordered word match would give the wrong answer
  because ICE_CREAM is treated as CREAM_ICE, but the other two words are
  left as-is. Not a perfect example, since the existing match function would
  give the correct answer, but this becomes more relevant with higher word
  counts).
- ColorListPresets of the form 'JUST_colourname' are now dynamically
  generated from all colours matching 'CLOTHING_colourname', as long
  as they aren't already explicitly defined.
- As a part of the above point, JUST_DARK_RED has been renamed to
  JUST_RED_DARK. All existing mods that use this preset will be
  functionally unaffected, due to the unordered-word matching, but
  the error.log will include a message for each instance. All
  instances in the repo of JUST_DARK_RED and JUST_DARK_BROWN have
  been updated to the correct new form.
- Added additional details to the closest-match warnings to describe
  what the difference between the input and its nearest match is.
- getClothingTypeFromId() now accepts a second slotHint parameter.
  If this parameter matches the name of an InventorySlot, the
  returned clothing type will be restricted to only clothing items
  that can occupy that slot. Currently only utilised for loading
  equipped clothing from XML.

### Testing
- loading my own game and saves, with additional instrumentation
- attempting several different incorrect values for mods and similar xmls


Innoxia:
This looks really interesting! I'm a little too tired to understand it fully at the moment, but I'll read over it and most likely merge it tomorrow!

src/com/lilithsthrone/game/inventory/clothing/ClothingType.java
```diff
@@ -6358,13 +6359,35 @@ public String unequipText(GameCharacter clothingOwner, GameCharacter clothingRem
...
+	public static AbstractClothingType getClothingTypeFromId(String id, String slotHint) {
 //		System.out.print("ID: "+id);
 		
 		if(oldIdConversionMap.containsKey(id)) {
 			id = oldIdConversionMap.get(id);
 		}
 		
-		id = Util.getClosestStringMatch(id, idToClothingMap.keySet());
+		Map<String, AbstractClothingType> choiceMap = idToClothingMap;
+		if (slotHint != null) {
```
Innoxia, line 6373: I've changed this in my local code to:
`if (slotHint!=null && !slotHint.isEmpty()) {`

Without the empty String check, my error log was being spammed with warnings about InventorySlot not having a valid value for an empty String.

CognitiveMist: Interesting, I hadn't seen that myself, that I remember.

PR #1370, opened 2020-07-24
Offspring can inherit generic wings and horns

Stadler76:
If the father and/or the mother has generic (feathered, leathery or insect) wings the offspring can inherit them.
**Update:** Now this applies to generic horns, too.

  Made sure that arm wings will block back wings and if the wing types differ it depends on whom the child takes after what wings it gets or no wings if only one parent has wings.
  So this results in a chance that the offspring of a wolf-morph and a pegasus will be winged wolf-morphs.
  As the game has now generic horns as well I've updated this PR to include inheritance of generic horns.

  Note: While I could use
  ```java
  public boolean hasGenericWings() {
  	return this.body.getWingType().getRace() == Race.NONE && this.hasWings();
  }
  ```
  respectively
  ```java
  public boolean hasGenericHorns() {
  	return this.body.getHornType().getRace() == Race.NONE && this.hasHorns();
  }
  ```
  I decided against that as I prefer 'tagging' it with:
  ```java
  @Override
  public boolean isGeneric() {
  	return true;
  }
  ```

No

Yes, 0.3.8.9 Alpha and up to 0.3.15 Alpha

`@Stadler#3007`


Stadler76:
#### Added a couple more reusable methods
`chonker++` -> Added the following reusable methods for wings:
- AbstractWingType and WingType: `isGeneric()`
- Wing: `setType(AbstractWingType type)`, `setSize(int wingSize)`, `setTypeAndSize(AbstractWingType type, int wingSize)`
- GameCharacter: `hasGenericWings()`

Innoxia:
Thank you for this!

PR #1371, opened 2020-07-24
Search in Watering Hole based on preferences

AceXP:

Searching for a partner in the Watering Hole will no longer show genders or subspecies that cannot normally be spawned (based on the gender/subspecies preferences).

Added checks to see if the GenderPreference or Subspecies(Masculine or Feminine)Preference is greater than zero. If it isn't these genders and subspecies will not be selectable anymore. Note: Humans are always selectable, since there is no SubspeciesPreference for Humans.

No

Tested in 0.3.8.9


Innoxia:
Thank you for this PR, but I think that it might be better leaving the options for otherwise-disabled NPC genders/races available. Maybe some players don't want to encounter some races as random encounters (and so disabled them), but would be ok with sometimes selecting them in the club?
I'm not entirely 100% sure on this, so I'll leave the PR open for if anyone wants to give their input.

AceXP:
Taking away options might indeed not be for the best, so I will close this now. It's not that you _have_ to push those buttons, even if they are there.

Issue #1369, opened 2020-07-23
Taking Inventory From Arcane Arts


EsHoshizora:
When enchanting items while trading with Vicky doing the same enchantment until you run out of items will put you back on the inventory trading screen, but will allow you to use the "take all" button to take everything in Vicky's inventory. This works for both the "weapons" and the "spells & potions" inventories, but was not tested on other shopkeeps.

Taking all of Vicky's items was unexpected, as the button should have stayed grayed out during the trading interactions.

1. Go to Ralph's Snacks and purchase an amount of an enchantable item (tested it with several of the transforming foods) or otherwise have a quantity on hand
2. Go to Arcane Arts and begin trading with Vicky and open either the weapons or spells & potions inventory menus
3. Select one of the enchantable items and open the enchanting menu by clicking the button in the corner
4. Select the random enchantment and select craft
5. Continue selecting craft until the item is all used up
6. without moving the mouse from that position, click again when the normal trading inventory screen reappears (the take all button is in the same space as the craft button)
7. The error then occurs, allowing you to take all of Vicky's inventory (this works on both weapons and spells & potions)

Error Code?:

Failed to load character present home: 14,DominionAlleywayAttacker
Failed to load character present home: 13,DominionAlleywayAttacker
Failed to load character present home: 12,DominionAlleywayAttacker
Failed to load character present home: 10,DominionAlleywayAttacker
Failed to load character present home: 11,DominionAlleywayAttacker
java.lang.OutOfMemoryError: Java heap space
	at java.awt.image.DataBufferInt.<init>(Unknown Source)
	at java.awt.image.Raster.createPackedRaster(Unknown Source)
	at java.awt.image.DirectColorModel.createCompatibleWritableRaster(Unknown Source)
	at java.awt.image.BufferedImage.<init>(Unknown Source)
	at com.lilithsthrone.rendering.CachedImage.scaleDown(CachedImage.java:131)
	at com.lilithsthrone.rendering.CachedImage.load(CachedImage.java:35)
	at com.lilithsthrone.rendering.ImageCache.getImage(ImageCache.java:98)
	at com.lilithsthrone.game.character.GameCharacter.getCharacterInformationScreen(GameCharacter.java:3113)
	at com.lilithsthrone.game.dialogue.utils.CharactersPresentDialogue.resetContent(CharactersPresentDialogue.java:42)
[...]

PR #1372, opened 2020-07-25
Disallow enchanting while trading

AceXP:

Fix the issue where returning from the Enchantment dialogue gives you control over the shopkeepers inventory (#1369) 

Disallow enchanting items while trading

No 

Yes, tested in 0.3.8.9


Innoxia:
I've already fixed this on my end, but that you for the PR regardless!

PR #1375, opened 2020-08-03
Fix minor issues in dialogues

AceXP:

Fix minor text issues in dialogues

In the Enforcer Alleyway dialogue the reference should be to the npc3 (character being enslaved), not to the Enforcer that is present

Fixed some typos in the race info text

No

Yes, tested in 0.3.8.9


Innoxia:
Thank you for these fixes!

PR #1384, opened 2020-08-11
Check all arms for 2-h weapons

AceXP:

Fix a bug where having multiple arm pairs allows an all-out strike with a 2-handed weapon

Check all arm pairs for 2-handed weapons. Only if at least one pair is not using 2-h weapons is an all-out strike allowed. Also check if the main weapon is 2-handed, if so, do not apply any damage in the off-hand attack.

No

Yes, tested in version 3.9.1 with various weapon combinations and arm pairs.

AceXp#0930


Innoxia:
Thank you for this!

PR #1386, opened 2020-08-11
No prude succubus alleyway attackers

AceXP:

Prevent any succubus alleyway attackers from spawning with the 'Prude' trait, as it doesn't fit them.

Remove the trait (if present, it's a very low chance anyway)

No

Yes, version 0.3.9.1

AceXp#0930

### DIFF
--- a/src/com/lilithsthrone/game/dialogue/npcDialogue/dominion/AlleywayDemonDialogue.java
+++ b/src/com/lilithsthrone/game/dialogue/npcDialogue/dominion/AlleywayDemonDialogue.java
@@ -8,6 +8,10 @@ import com.lilithsthrone.game.character.GameCharacter;
 import com.lilithsthrone.game.character.attributes.AffectionLevel;
 import com.lilithsthrone.game.character.attributes.CorruptionLevel;
 import com.lilithsthrone.game.character.body.CoverableArea;
+import com.lilithsthrone.game.character.body.types.PenisType;
+import com.lilithsthrone.game.character.body.valueEnums.PenetrationGirth;
+import com.lilithsthrone.game.character.body.valueEnums.PenetrationModifier;
+import com.lilithsthrone.game.character.body.valueEnums.TesticleSize;
 import com.lilithsthrone.game.character.fetishes.Fetish;
 import com.lilithsthrone.game.character.npc.NPC;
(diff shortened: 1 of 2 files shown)


src/com/lilithsthrone/game/dialogue/npcDialogue/dominion/AlleywayDemonDialogue.java
```diff
@@ -707,7 +707,7 @@ public Response getResponse(int responseTab, int index) {
 			return null;
 		}
 	};
-	
+
```
CognitiveMist, line 710: Since you reverted the penis-growth part of your change, this file is only whitespace changes. It is usually considered good git manners to not do whitespace-only changes to files, since that makes merging trickier with no real benefit.

There are a couple ways to accomplish this; normally reverting changes is done with `git revert`, in this case you might run something like `git revert 1f2462b 56db55d`. This command will commit the reversion immediately; to look over the changes first, use the `--no-commit` flag.

Alternatively, you could just grab a known-clean version of that specific file, with `git checkout`. The exact command you'd use here would depend on how exactly you've configured your remotes; for me, it would look like `git checkout innoxia/dev src/com/lilithsthrone/game/dialogue/npcDialogue/dominion/AlleywayDemonDialogue.java`, since I've called Innoxia's upstream repo `innoxia`. You may not have added Innoxia's repo as a remote, in which case you could use `git checkout dev src/com/...` (assuming the file in your dev branch is untouched from Innoxia's version) or just specify a commit before you performed those changes, like `git checkout b8d4fb0 src/com/...`.

I think both these methods are good to know, the first is more of a sledgehammer, while the second is more akin to a scalpel. But since your changes are just a single file, they work equally well here.

I hope I'm not sounding too preachy, just trying to share the things I've learned over the years.

AceXP: I will do a proper revert later, it was late, I was tired and not thinking straight. And working with git is still new to me so I appreciate the help.

AceXP: Did a checkout back to the original version as advised.

Innoxia:
Thank you!

PR #1387, opened 2020-08-12
No enchanting in trade

AceXP:

Fix the issue where returning from the Enchantment dialogue gives you control over the shopkeepers inventory (#1369)

Disallow enchanting items while trading

No

Yes, tested in 0.3.8.9

@Innoxia Made a new PR for this


AceXP:
Need to check if enchanting tattoos still works after this.

AceXP:
No longer needed, will be fixed in new version.

PR #1388, opened 2020-08-12
Use correct cutoff point for arcane impotence

### LINKED ISSUE #1374: [0.3.8.9] Aura is calculated incorrectly
**Describe the bug**
The Maximum Aura is calculated incorrectly at low Arcane levels. A character's Arcane stat and level are not taken into the calculations for Maximum Aura when the Arcane stat is between 10 and 14. Likely a remnant from earlier game versions where Arcane levels (Impotence, Potential, Proficiency, etc.) had different numerical ranges than in the present version. In an earlier version, Arcane Potential had a range of 5-15 and included an "Aura limited to 5" effect, which in the current version should only apply to Arcane Impotence.

**Expected behavior**
Maximum Aura should be calculated based on a characters Arcane stat in the following way:

Arcane < 10
Maximum Aura set to 5

Arcane 10 or higher
Maximum Aura is calculated according to 5 + (2*level) + (5*Arcane) + Bonus Aura

How it is actually calculated in game:

Arcane < 10
Maximum Aura set to 5

Arcane between 10 and 14
Maximum Aura is calculated according to 5 + Bonus Aura

Arcane 15 or higher
Maximum Aura is calculated according to 5 + (2*level) + (5*Arcane) + Bonus Aura

**To Reproduce**
1. Set a character's Arcane stat to a value between 10 and 14 inclusive
2. Observe the resulting Maximum Aura

AceXP:

Arcane impotence should apply when arcane is lower than 10, but it is applied when arcane is lower than 15

Change the cutoff value to 10

No

Yes, tested in version 0.3.9.2

AceXp#0930

Closes #1374

### DIFF
--- a/src/com/lilithsthrone/game/character/GameCharacter.java
+++ b/src/com/lilithsthrone/game/character/GameCharacter.java
@@ -5542,7 +5542,7 @@ public abstract class GameCharacter implements XMLSaving {
 		}
  		// Special case for mana:
 		if (attribute == Attribute.MANA_MAXIMUM) {
-			if(getAttributeValue(Attribute.MAJOR_ARCANE) < 15) {
+			if(getAttributeValue(Attribute.MAJOR_ARCANE) < 10) {
 				value = 5;
 			} else {
 				value = 5 + 2*getLevel() + 5*getAttributeValue(Attribute.MAJOR_ARCANE);


Innoxia:
Thank you for this, but I decided to replace the;
`getAttributeValue(Attribute.MAJOR_ARCANE) < 15`
with
`getAttributeValue(Attribute.MAJOR_ARCANE) < IntelligenceLevel.ONE_AVERAGE.getMinimumValue()`
so if I ever change the boundaries for it again in the future, it will be covered b this.

Innoxia:
(I'll credit you for this fix regardless, as I wasn't aware of this issue before seeing how you fixed it.)

AceXP:
@Innoxia Way better solution, as magic numbers are never a good idea.

PR #1389, opened 2020-08-13
Encounter standardization

DSGMods:
To standardize all encounters to use getDialogues()

Changed SUBMISSION_TUNNELS, DOMINION_DARK_ALLEY, and BAT_CAVERN to override getDialogues() instead of accepting Util.newHashMapOfValues() as a parameter

No

0.3.9.2
All encounters still work, verified again after clean build

dsg#4931


Innoxia:
Thank you for this!
(I took things one step further and removed the 'dialogues' HashMap from the Encounter constructor entirely.)

PR #1391, opened 2020-08-13
Add new option for sex action bypass

AceXP:

Add a new option for the bypass sex action content setting.

Remove the boolean value, and add a property. There are three possible values: None, Limited and Full:

None: There will be no options to bypass sex action corruption requirements, you are limited in your actions based on your corruption and fetishes. This is the same as the old OFF

Limited: Sex action corruption requirements may be bypassed if your corruption level is one level below the required corruption level of the action, but you will gain corruption if you do so.

Full: All sex action corruption requirements may be bypassed, but you will gain corruption if you do so. This is the same as the old ON.

No

Yes, tested in version 0.3.9.2

AceXp#0930

@CognitiveMist Any chance that you take a look at these changes? I did extensive testing, but in this case the 4 eye principle applies.

Bypassoptions

### DIFF
--- a/src/com/lilithsthrone/controller/MainControllerInitMethod.java
+++ b/src/com/lilithsthrone/controller/MainControllerInitMethod.java
@@ -6179,6 +6179,14 @@ public class MainControllerInitMethod {
 					setAutosavePreference(id, i);
 				}
 			}
+
+			// bypass sex action options:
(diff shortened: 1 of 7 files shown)


Innoxia:
Thank you for this!

PR #1392, opened 2020-08-16
Distinguish between bestial and non bestial fluids for addictions.

### LINKED ISSUE #1364: Addiction bug.
Sorry for bad english, i'm from Russia.
**Describe the bug**
If you consume some kind of demonic addictive fluid, game will say, what you become addicted to demonic horse fluid. It doesn't matter who actually the liquid belongs to. I tested this bug with demonic cat, dog, leopard, harpy and succubus/incubus in alleyways. No matter what kind of liquid you consume, you always become addictive to demonic horse cum, girlcum of milk.
**Expected behavior**
I expected the correct description that you become addicted to a particular half-demon or demon fluids, not just always demonic horse.

**To Reproduce**
Just consume demonic addictive fluid from any NPC.

**Additional information:**
 - Game version 0.3.8.9
 - OS: Windows 7
 - Using .exe file
image
image
image

AceXP:

Determine if an addictive fluid is from a bestial or a non bestial source, and store this information.

Add a new property to addictions, save it to xml, load it from xml and initialize it from the fluid that is causing the addiction. And finally show the proper name in the addiction status effect.

No

Yes, tested with version 0.3.9.2

AceXp#0930

Closes #1364

### DIFF
--- a/src/com/lilithsthrone/game/character/GameCharacter.java
+++ b/src/com/lilithsthrone/game/character/GameCharacter.java
@@ -17629,7 +17629,7 @@ public abstract class GameCharacter implements XMLSaving {
 		for(Entry<SexAreaOrifice, List<FluidStored>> entry : this.fluidsStoredMap.entrySet()) {
 			for(FluidStored fs : entry.getValue()) {
 				if(fs.getFluid().getFluidModifiers().contains(FluidModifier.ADDICTIVE)) {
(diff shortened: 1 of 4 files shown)


AceXP:
@Innoxia As you said: I think I won't make a distinction between feral/taur/normal fluid addictions

PR can be closed. All fluids will always be regular, never bestial. The display of the addiction will be fixed.

PR #1403, opened 2020-08-30
Reapply Heavy Makeup

CognitiveMist:
While carrying an arcane makeup set, characters will reapply heavy lipstick whenever it wears off.

image

Works for the player and for npc's.

As a part of this change, added an overload to `AbstractItemType.getName` and `AbstractItemType.getNamePlural` that allows specifying whether the result should be capitalized.


Innoxia:
Thank you for this! It's a really good quality of life change.

PR #1404, opened 2020-08-31
Fix a parser error in the vagina reveal description

AceXP:

Fix a parser error in the vagina reveal description

-- Switched the parser targets (revealer and revealed) to make it compatible with the penis reveal description
-- Do the parsing at the end, rather than in each seperate case
-- And thereby fix the bug in the (Fetish.FETISH_VAGINAL_GIVING).isNegative()) part

No

Yes, tested in 0.3.9.5

AceXp#0930

### DIFF
--- a/src/com/lilithsthrone/game/character/GameCharacter.java
+++ b/src/com/lilithsthrone/game/character/GameCharacter.java
@@ -14071,7 +14071,7 @@ public abstract class GameCharacter implements XMLSaving {
 			if(reactingPace!=SexPace.SUB_RESISTING
 					&& characterReacting.getFetishDesire(Fetish.FETISH_VAGINAL_GIVING).isNegative()) {
 				sb.append(UtilText.returnStringAtRandom(
-						"[npc.Name] tuts in disappointment as [npc2.namePos] [npc2.pussy+] is revealed, ",
+						"[npc.Name] [npc.verb(tut)] in disappointment as [npc2.namePos] [npc2.pussy+] is revealed, ",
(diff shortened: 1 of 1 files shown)


src/com/lilithsthrone/game/character/GameCharacter.java
```diff
@@ -14071,7 +14071,7 @@ public String getVaginaRevealDescription(GameCharacter characterBeingRevealed, G
 			if(reactingPace!=SexPace.SUB_RESISTING
 					&& characterReacting.getFetishDesire(Fetish.FETISH_VAGINAL_GIVING).isNegative()) {
 				sb.append(UtilText.returnStringAtRandom(
-						"[npc.Name] tuts in disappointment as [npc2.namePos] [npc2.pussy+] is revealed, ",
+						"[npc.Name] [npc.verb(tut)] in disappointment as [npc2.namePos] [npc2.pussy+] is revealed, ",
```
CognitiveMist, line 14074: There is also a similar `tuts` that should be fixed in the penis reveal function... maybe others too?

CognitiveMist: Actually, the more I look in this file, the more I find. It could end up being a chonker of a PR, so maybe don't worry about it right now.

AceXP: It's a good point, but I am not even sure this particular bit can be triggered with the MC as the tutting character. I tried to get this but I did not succeed so far.

Innoxia:
Thank you for this!

PR #1406, opened 2020-09-04
Fixed time bug in Angel's Kiss Sub scene

void-weaver:
Not a large pull request, but I can include these things.

Fixes a balance/continuity issue when acting as a submissive partner in the Angel's Kiss

Dialogue states that the client enters "Around twenty-five minutes later", but the ANGELS_KISS_SELL_SELF_SUB DialogueNode only progresses time by 5 minutes. This has been fixed to make the node take 25 minutes instead.

No.

Only so far as to ensure that the fix works, on version 0.3.9.7


Innoxia:
Thank you for this!

PR #1407, opened 2020-09-05
Fixed Sean's fight scene

void-weaver:

Fixes an issue where Accepting Sean's challenge in the Slaver's alley wouldn't initiate a fight scene

In commit 3ec95aa, the NPC's base clothing was changed without changing the fight scene preparation, wherein Sean removes some equipment, including a tie that no longer exists. This lead to a NullPointerException in `GameCharacter.unequipClothingIntoInventory`, which prevented the scene from moving any further into the fight sequence.  
This PR changes the fight scene preparation.

No.

Tested in 0.3.9.7, only so far as to ensure that the fix works

Null#3561

-----

Being largely unfamiliar with the codebase, I am unaware of any other instances of this issue, but they are possible. To prevent this issue from affecting the game, a check can be made to return early in `GameCharacter.unequipClothingIntoInventory` if the clothing parameter is null.


Innoxia:
Good catch, thank you for this!

PR #1411, opened 2020-09-14
Player Selfie uses custom description if available

aDrunkLittleDerp:
The player's description always shows the default description ("Having been pulled into an enchanted mirror..."), even if changed with save editing. This change makes it use the default description if there's no custom description.

Tested with 3 new characters, 1 character with custom description, and 1 pre-existing character, on 0.3.9.8.


Innoxia:
Thank you for this! I've got it merged for the next update!

PR #1412, opened 2020-09-15
Fix anus description of elementals

AceXP:

Fix the issue where the description of the anus of elementals is wrong

Added coverings for elementals that have the required pattern 'ORIFICE_ANUS', and create these with the CreateElemental BodyCoveringTemplateFactory. Also changed this method by adding a parameter 'pattern'. Then use these new patterns as needed in GameCharacter.getCovering. And changed the covering modifier to 'glittering' for ice, because I think it's cool...

No

Yes, tested with version 0.3.9.8

AceXp#0930

Info screen before the fix:
BeforeFireElemental

Info screen after the fix:
AfterFireElemental

The human/demonic description is based on what the elemental has, and is not affected by this fix.

### DIFF
--- a/src/com/lilithsthrone/game/character/GameCharacter.java
+++ b/src/com/lilithsthrone/game/character/GameCharacter.java
@@ -20455,7 +20455,7 @@ public abstract class GameCharacter implements XMLSaving {
 		clothing.onUnequipApplyEffects(this, characterRemovingClothing, false);
 		inventory.forceUnequipClothingIntoVoid(this, characterRemovingClothing, clothing);
 	}
-	
+
 	public String unequipClothingIntoVoid(AbstractClothing clothing, boolean automaticClothingManagement, GameCharacter characterClothingUnequipper) { // TODO it's saying "added to inventory"
 		boolean unknownPenis = !this.isAreaKnownByCharacter(CoverableArea.PENIS, Main.game.getPlayer()) && !this.isCoverableAreaVisible(CoverableArea.PENIS);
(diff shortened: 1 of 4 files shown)


CognitiveMist (commented):
Do you know why this is needed for ANUS, but not PENIS, MOUTH, NIPPLES, NIPPLES_CROTCH, and VAGINA?

AceXP:
Why this is only needed for ANUS is a bit of a mystery. I did wonder about it myself, but have not found an answer.

AceXP:
Due to the extensive changes in BodyCovering, this needs to be revised. However, maybe there is a simpler solution to the underlying issue. Will keep it as a draft for the moment.

Innoxia:
I've got this fixed in my local code, thank you for bringing my attention to this issue!

The problem was that elementals had a new `AbstractBodyCoveringType` generated for each `BodyCoveringCategory` (so that I wouldn't have to define every single covering type for every non-flesh material). These automatically-generated covering types used the `CoveringPattern` `NONE` as their pattern. As the pattern is what is responsible for generating covering descriptions, the description for all orifices was the default description. This was affecting all orifices, but only the Anus showed a problem, as only the Anus used the parsing command `FullDescription` (as `[npc.anusFullDescription(true)]`), so that was the only part which could be identified as having the underlying issue.

I've fixed this by adding a switch on the BodyCoveringCategory when generating elemental coverings (in `BodyCoveringType.java`), inside of which I manually set the pattern for the automatically-generated AbstractBodyCoveringType to the correct orifice pattern for all orifices (I also did the same for eye coverings).

AceXP:
Closing now, as it was solved (and much more elegantly).

PR #1413, opened 2020-09-20
Elasticity affect depth

Lightcanadian:

I didn't like that you had to have a cavernous pussy(or other orifices) in order to accommodate a longer dick so I attached all the orifice to elasticity since I felt that it could make sense that an elastic orifice can accommodate not only larger but longer penetrative object

1. Added an ON/OFF setting to enable the feature(default off)
1. Change all the orifice uncomfortable value to return different depth depending on the elasticity with a minimum value of the default value, also set the modifier value to something that I felt was not overly exaggerated

none

Tested in 0.3.9.9

LightCanadian

### DIFF
--- a/src/com/lilithsthrone/controller/MainControllerInitMethod.java
+++ b/src/com/lilithsthrone/controller/MainControllerInitMethod.java
@@ -6286,6 +6286,7 @@ public class MainControllerInitMethod {
 					new Value<>("ANAL", PropertyValue.analContent),
 					new Value<>("GAPE", PropertyValue.gapeContent),
 					new Value<>("PENETRATION_LIMITATION", PropertyValue.penetrationLimitations),
+					new Value<>("PENETRATION_LIMITATION_DYNAMIC", PropertyValue.elasticityAffectDepth),
 					new Value<>("FOOT", PropertyValue.footContent),
 					new Value<>("FUTA_BALLS", PropertyValue.futanariTesticles),
 					new Value<>("CLOACA", PropertyValue.bipedalCloaca),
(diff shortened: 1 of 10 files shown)


AceXP:
This is an interesting idea, but I think it is far from complete.

Basically, what you added is that an elastic orifice can be stretched lengthwise and accommodate more length. How quickly does this happen? What is the effect on the owner of the orifice? How quickly will the orifice recover? All these factors are in the current stretching effects, but there are none of those here.

Also, minor issue, if you add a preference setting it should be saved to and loaded from the .xml file with preferences.

Lightcanadian:
@AceXP The property values are serialized, so it automatically adds or removes the property from the XML base on if it's enabled or disabled. You can see the change in the property.xml in the same location as the .exe.

Also, most of what you mention is already present in the game, it was just not affecting depth. The only thing that I could think about is a flavor text of how you are stretch with maybe some reaction.

AceXP:
I don't think any automatic serialization is done at the moment. It's done by savePropertiesAsXML in properties.java

You are right that there is a distinction between stretching (diameter too big) and uncomfortable insertions (length too long). With this change, the maximum depth is increased, but only for uncomfortable insertions, so all those effects will still apply. I just feel that using that extra depth should maybe have an additional effect as well? Something is being stretched. but in another direction. A very long and thin object will be uncomfortable, but it will not be stretching anything.

Lightcanadian:
If you go in MainControllerMethod.java, I have added PENETRATION_LIMITATION_DYNAMIC in the setting map (maybe I should change the name), the setting map handle toggle setting by itself and add or remove a flag in properties.xml. When enable, the line 
`<propertyValue value="elasticityAffectDepth"/>`
will appear under 
`<propertyValues>`
Otherwise(if disable) the line will just be missing and considered disable.
Therefore the setting is saved as XML and is tested and working

AceXP:
Well, it looks like these simple properties are automatically saved because savePropertiesAsXML loops over all of them. So that is why your new property is saved as well. But this doesn't hold true for all types of properties.

Innoxia:
Thank you for this!

PR #1415, opened 2020-09-21
Save character list for each call when parsing

AceXP:

Fix the bug where characters get lost when parsing speech, leading to parserTarget.Character() not found! errors.

Save the current list of characters for parsing, and restore it when the parsing is done, so that calling parse recursively works properly

No

Yes, tested with 3.9.9.9

AceXP#0930


Innoxia:
Thank you for this!

PR #1419, opened 2020-09-25
Fix bug when equipping multiple clothing items

AceXP:

Fix a bug when multiple identical clothing items are equipped in succession.

Changed the equipClothingOverride method. If a clothing item to be equipped (newClothing) has multiple occurances (i.e. the count of the item is greater than one) equipping this item will change the key of the item in the map where the item is stored, and thus change all other occurances of that item. This has consequences and leads to bugs with the next occurances of the item.

This is the situation where this happens:

>go to my room after a long day of having nothing to do in dominion
>decide to have a bath with my bedwarming slaves
>have quick but passionate impregnation post bath sex with all of my slaves
>notice that all but one of them don't have their uniform on them
>try to pick them up but my inventory isn't opening

The slaves have the same clothing, so when entering the bath all these are deposited on the floor, with a count. When after the bath the first character has the items restored, this also changes the other items still on the floor. This means that a) the other characters will not get dressed as the item they need can not be found anymore. And b) the inventory will break because there are now items on the floor that are equipped to a slot.

Also, when automatic replacement of clothing is off, there will be an attempt to pick up the clothing from the floor. This must not be done if the clothing is worn already, as in the case when there are multiple instances on the floor, the first characters will get one worn as well as one picked up, leaving the later characters with no clothing at all.

No

Yes, tested with 3.9.9

This PR is the work of deboucher#8190 (submitted byAceXP#0930)


CognitiveMist (approved):

Innoxia:
Thank you for this!

PR #1420, opened 2020-09-26
Disallow access to equipped items while enchanting

AceXP:

Fix the bug where clicking on an equipped item while enchanting during a trading session gives you full access to the inventory of the trader, allowing to take all their equipped items and the items in their inventory.

The InventorySelectedItemEventListener will do nothing when you are in the ENCHANTMENT_MENU. I.e. when you now click an equipped item while enchanting, nothing will happen.
Since you can never enchant equipped items, this action is not needed anyway.

No

Yes, tested with 3.9.9

AceXP#0930


CognitiveMist (approved):

Innoxia:
Thank you for this!

PR #1422, opened 2020-09-27
Added check for fertility pills, including broodmother

AceXP:

Special NPC's (like Lilaya and Kate) should treat brood mother pills much the same as other fertility pills.

Added a isTypeOneOf method to AbstractItem, and use this when determining if a special reaction is needed.

Also, Roxy now sells brood mother pills, and there is a chance random NPC's will have them as well.

And several scenes still mention the Breeder pills as pink, but they have been purple for some time. Those occurances have been changed.

No

Yes, tested with 3.9.9

AceXP#0930

### DIFF
--- a/src/com/lilithsthrone/game/inventory/item/AbstractItem.java
+++ b/src/com/lilithsthrone/game/inventory/item/AbstractItem.java
@@ -4,8 +4,8 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Stream;
 
-import com.lilithsthrone.game.inventory.enchanting.TFModifier;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.NodeList;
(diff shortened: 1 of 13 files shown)


src/com/lilithsthrone/game/inventory/item/AbstractItem.java
```diff
@@ -282,6 +283,11 @@ public boolean isGift() {
 		return itemType.isGift();
 	}
 
+	public boolean isFertilityPill() {
+		return this.getItemType().equals(ItemType.getItemTypeFromId("innoxia_pills_fertility"))
+				|| this.getItemType().equals(ItemType.getItemTypeFromId("innoxia_pills_broodmother"));
+	}
+
```
CognitiveMist, line 290: I wonder if a more generic `item.isTypeOneOf(String itemType...)` would be better. More useful for modders, I think.

src/com/lilithsthrone/game/inventory/item/AbstractItem.java
```diff
@@ -5,6 +5,7 @@
 import java.util.List;
 import java.util.Set;
 
+import com.lilithsthrone.game.inventory.enchanting.TFModifier;
```
CognitiveMist, line 8: Extra?

src/com/lilithsthrone/game/inventory/item/AbstractItem.java
AceXP: Done

src/com/lilithsthrone/game/inventory/item/AbstractItem.java
AceXP: And gone

CognitiveMist (approved):

Innoxia:
Thank you for this!

PR #1423, opened 2020-09-28
Enchantment recipes can now be applied to weapons that already have enchantments

AceXP:

Make it possible to apply recipes (stored enchantment lists) to weapons that already have an enchantment

Choose a weapon with no enchantment to apply the recipe to, and if none are available choose the first one that has (one or more) enchantments. In all cases, the weapon must be the same type as the one the recipe was made for.

Also, when loading a recipe, add the effects on the item that is to be enchanted to the list of effects in the enchantment dialogue, and not just the list of effects that were in the recipe.

No

Yes, tested with 3.9.9

AceXP#0930


Innoxia:
Thank you for this!

Issue #1424, opened 2020-09-29
Milking Room Only Allows 7 Slaves At A Time


GreatSwordOfVictory:
The Milking Room description says that up to 8 slaves can be using it at a time, but in game you can only assign 7 slaves at a time. Might be because the game saves one of the "spots" for player use? Because 7 slaves + 1 player = 8 total? So I wasn't sure if it was a bug or a error in the room's description text, but figured I'd point it out here.

AceXP:
It would seem to be a bug, as the intention seems to be to disallow the player the use of the room if it is fully occupied with 8 slaves.

AceXP:
I've looked into this, and I think that everything works as expected. You can assign 8 slaves to each milking room. It's just that AFTER you've assigned the eight slave, the time slots for that slave will grey out for the Dairy Cow job, and the tooltip will say 'not enough space in milking rooms'. But that is logical, since this slave is now occupying the last available slot, so it's not available for another slave. But the slot DOES show the yellow color for that job, and the slave IS assigned to it at that time. Also, when you visit the milking room at an appropriate time, all the eight assigned slaves will be there.
I had 7 slaves on 24hr duty, and three other slaves each working 8 hours and that works just fine. I am sure other combinations will work too.
Could you check this and close this issue if you agree?

GreatSwordOfVictory:
That's what I'm seeing too, glad I'm not going crazy lol. Not sure if the 8th slave assigned is having their fluids milked or not. I would check right now but I have to run and do IRL stuff. If nobody else can verify before I get back, I'll try poking around the issue more to see if I can find anything else out.

GreatSwordOfVictory:
Alright, I think AceXP got it right when he said something funny is going on. So, you can assign 8 slaves to be milked. That is working properly. The thing is, only 7 slaves are in the milking room during their assigned times, the whole time. There's some weirdness going on with the 8th slave, and what I saw was that at the first hour, every slave went in. At the second hour, the 8th slave left and went back to their room, where they had the milking stuff equipped like they were in the milking room. The good news is that the game seems to still track the fluids being milked properly, so you don't loose out on any money or whatever.

I have no idea what kind of black magic makes this game work, but to my baby programmer brain, this makes me think something's wrong with how the milking room handles having more than 8 slaves or something, making it boot the 8th slave back to their room when it's not supposed to. It's weird, haha.

GreatSwordOfVictory:
[...]

PR #1425, opened 2020-10-03
Update handling of slave jobs and loitering

AceXP:

Allow the full complement of slaves (8) to be in a milking room at the same time.

Fix the (minor) issue where slaves on cleaning duty gather around the MC when loitering in the corridor in Lilaya's home.

Added an option to rest for eight hours in your room.

Fix the description for a room with a single steel bed showing the text for a comfortable double size bed.

Bug: If you sleep for longer than the 'well rested' duration, the effect is there but not applied anymore. This has been fixed.

When checking to send a slave to a milking room, don't check for a free slot (as there should always be room since the maximum number of allowed slaves is determined earlier).

Also changed the selection of a free slot in a milking room, based on capacity and preferred type.

Also changed the order in which slaves are assigned new jobs, those becoming idle will be moved first.

Also introduced a new state for the MC called isActive. The MC becomes inactive when sleeping or loitering (or walking a path). When inactive, all slaves are considered free to go anywhere.

When sleeping, first apply the well rested effect (but now including the sleep duration) and only then end the turn.

No

Yes, tested with 3.9.9

AceXP#0930

This PR closes issue #1424 Milking Room Only Allows 7 Slaves At A Time

Loitering


src/com/lilithsthrone/game/character/PlayerCharacter.java
```diff
@@ -1649,6 +1651,14 @@ public boolean isDiscoveredWorldMap() {
 		return this.isQuestProgressGreaterThan(QuestLine.MAIN, Quest.MAIN_2_D_MEETING_A_LILIN);
 	}
 
+	public boolean isInActive() {
+		return !isActive;
+	}
+
+	public void setActive(boolean active) {
+		isActive = active;
+	}
```
bicobus, line 1660: Can't `setActive()` have a default `True` boolean, as the name suggest?

Can't people use `player.isActive = False` instead?

src/com/lilithsthrone/game/character/PlayerCharacter.java
```diff
@@ -1649,6 +1651,14 @@ public boolean isDiscoveredWorldMap() {
 		return this.isQuestProgressGreaterThan(QuestLine.MAIN, Quest.MAIN_2_D_MEETING_A_LILIN);
 	}
 
+	public boolean isInActive() {
+		return !isActive;
+	}
```
bicobus, line 1656: This is convoluted. if the `isActive` property is public, there is no reason to have this little condition bloating the code. Let the check be done when it has to be done.

bicobus (commented):
Made some nitpicks around the new active methods of the player class. I believe it can have a lesser footprint, so long it is properly documented.

src/com/lilithsthrone/game/character/PlayerCharacter.java
AceXP: That just entered my mind while I was doing something else, so yes, it needs a default value true. 

isActive has protected access, so you can only access in this class, or with this public method. It's fairly common to do it this way.

AceXP:
Having private/protected members and public access methods (getters and setters) is how it is (almost always) done. So I decided to stick to that.

src/com/lilithsthrone/game/dialogue/utils/PhoneDialogue.java
```diff
@@ -3072,75 +3072,66 @@ public Response getResponse(int responseTab, int index) {
...
-						Main.game.endTurn(60*60*12);
-						Main.game.setContent(new Response("", "", Main.game.getDefaultDialogue()));
+						Loiter(60*12);
 					}
 				};
 			}
 			return null;
 		}
 
+		private void Loiter(int minutes) {
```
CognitiveMist, line 3114: Minor nitpick: function names should start with a lowercase letter.

src/com/lilithsthrone/game/dialogue/utils/PhoneDialogue.java
```diff
@@ -3072,75 +3072,66 @@ public Response getResponse(int responseTab, int index) {
...
+				period = "fifteen minutes";
+			} else if (minutes == 60) {
+				period = "hour";
+			} else if (minutes == 60 * 4) {
+				period = "four hours";
+			} else if (minutes == 60 * 8) {
+				period = "eight hours";
+			} else if (minutes == 60 * 12) {
+				period = "twelve hours";
+			}
```
CognitiveMist, line 3126: This should probably be decoupled from the caller's values by using `Util.intToString(int)`. Something like this should work (don't worry about counting spaces... the eventual html interpretation will combine them):

```
String period = ""
int hours = minutes / 60;
int partialMinutes = minutes % 60;
if (hours != 0) {
    period += Util.intToString(hours) + " hours ";
}
if (partialMinutes != 0) {
    period += Util.intToString(partialMinutes) + " minutes "
}
if (minutes == 0) {
    period = "period "
}
```

src/com/lilithsthrone/game/dialogue/utils/PhoneDialogue.java
AceXP: Done

src/com/lilithsthrone/game/dialogue/utils/PhoneDialogue.java
AceXP: Done, including check to generate "You spend the next hour" instead of "You spend the next one hours".

CognitiveMist (approved):

Innoxia:
Thank you for this!

PR #1426, opened 2020-10-04
Change default permissions/job settings allocation

AceXP:

Prevent the job permissions of slaves in the stocks from being overruled

Allocating default permissions/job settings for slaves in the AddSlave method should only be done if the MC is the new owner of the slave. With this change, the settings for slaves in the stocks for Allow_oral, allow_anal and allow_vaginal stay the way they are when the slave is first generated.

Added a check to see if the slave has a vagina to the slaves in stocks to prevent any other male pregnancies in the future.

Also removed the getDescription for slaves in stocks, as this overrides the description as a 'slave in stocks' that they start with.

No

Yes, tested with 3.9.9

AceXP#0930

SlavesInStocks

Male pregnancies:
PregnantMale

Description before change:
before

Description after change:
after

### DIFF
--- a/src/com/lilithsthrone/game/character/GameCharacter.java
+++ b/src/com/lilithsthrone/game/character/GameCharacter.java
@@ -4599,22 +4599,24 @@ public abstract class GameCharacter implements XMLSaving {
 			}
 			slave.setOwner(this);
 			slave.setPendingClothingDressing(false);
-			
-			// Set up default permissions:
-			for(SlavePermission permission : SlavePermission.values()) {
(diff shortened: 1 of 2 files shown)


CognitiveMist (approved):

AceXP:
Version 3.10 already solves this, so the change in Gamecharacter.java is now obsolete. The other changes are not very critical.

Innoxia:
I ended up implementing this locally, and your fix in GameCharacter.java was still needed, although I ended up fixing it in a slightly different way. I've credited you in the patch notes for your fixes, and thank you for this!

PR #1438, opened 2020-10-21
Enhancement: Direct item parsing

CognitiveMist:

Add direct access to a core item (clothing, weapon or item) within the parser.

Should function in all CDATA fields within clothing, weapon, and item xmls. Does not work for statusEffects, since those are not directly tied to an item.

Examples:
`this [#item.getDamageType().getWeaponDescriptor()] riding crop` might give "this freezing riding crop".
`It is worth [#item.getValue()].` might give "It is worth 50".
```
#IF(item.isDirty())
	It is soaked through.
#ELSE
	It is dry.
#ENDIF
```
might give "It is soaked through."

This is functional for clothing/weapons that can be equipped in multiple slots, as well as when loose in the inventory.

No

Yes, on my own build of the game (3.9.9).

CognitiveMist


Innoxia:
I mentioned this to you on discord, but just to leave a note here about it as well: I tested this by putting `[item.getName(true, true)]` into an item description, and when hovering over it to see the tooltip, it threw some NullPointerExceptions (related to the item parsing target not having an associated character). I'll get back to this PR and try to get this issue (and any others) fixed at some point in the future.

CognitiveMist:
Hey, @Innoxia, finally getting back to this...

> I tested this by putting `[item.getName(true, true)]` into an item description, and when hovering over it to see the tooltip, it threw some NullPointerExceptions (related to the item parsing target not having an associated character).

So, the ITEM target intentionally throws a NullPointerException when ITEM.getCharacter(...) is called.... since there is no GameCharacter as a part of AbstractItem, and things that would use that info don't really make sense for an item: `[item.speech(I'm a potato)]`, `[item.she]`, `[item.ears]`.

I could have chosen the option to return Main.game.getPlayer() like STYLE does, and maybe that would be the right call (up to you, of course). Returning null would be another option, but that would disguise the issue: `[item.italic(...)]` would work, while `[item.speech(...)]` throws an exception.

Perhaps another option would be to throw a more specific error than NullPointerException that tells the modmaker what they're doing wrong. Which gets me to the main thrust of this:

`[item.command()]` is just wrong for the purpose. Perhaps we should prevent using that form at all.
`[#item.command()]` and `[##item.command()]` are the correct invocations.

The reason is: the first, original, form is GameCharacter-centric. The newer forms are not, they are just script commands.

A final option is to remove ITEM from ParserTarget entirely -- it doesn't actually need to be there. I had just added it for the purpose of documentation. This would display `INVALID_TARGET_NAME(item)` instead.


As a footnote, `[item.getName(true, true)]` is not even valid anyways. If the NPE is replaced with a return-null (or anything else), the result will be `command_unknown`. You probably meant either: `[item.name]` which will actually fail due to the lack of a GameCharacter, or `[#item.getName(true, true)]` which does return the item's name as expected.

Innoxia:
Thank you for this! I'm sorry that it took me so long, and you were right in your last comment that my testing syntax wasn't even valid - I think your PR was working fine from the start and it was just my error which made me not merge this sooner, so I'm very sorry for that.

PR #1439, opened 2020-10-22
Fix glow commands

CognitiveMist:
Fix glow commands

- bare `glow` was missing
- `glowFreedom` wasn't rainbow'd (and other rainbow-ables, too)

Tested on 0.3.9.9, no new assets required.

image


Innoxia:
Thank you for this!

PR #1442, opened 2020-11-02
Memoization for the script engine

CognitiveMist:
Increase performance by adding a memoization cache to compile scripting engine scripts

-- adds a cache intended to hold compiled forms of script engine scripts
-- cache capacity set to 500, and will stop adding new entries at that limit (my tests did not exceed 100, but mods affect this)
-- my tests showed scripting engine calls take 50% less time on average
-- WARNING: adds one more nashorn warning

No

Yes, I tested with Oracle JDK 14 and version 0.3.9.9
AceXP tested with Oracle JDK 8 and version 0.3.9.9

The Cognitive Mist (Phlarx#1765)

- **If you want quick feedback, you can use @Innoxia, or jump on the Lilith's Throne discord and send Innoxia a PM.**


Innoxia:
Thank you for this!

Issue #1444, opened 2020-11-04
Empty Tooltips


InfinitePondera:
All tooltips are empty when you hover over buttons, items, etc. instead of showing whatever text and/or image it was supposed to show.

![image](https://user-images.githubusercontent.com/28277461/98092498-819cd000-1e65-11eb-9fb9-000ede834dc1.png)

![image](https://user-images.githubusercontent.com/28277461/98092534-8a8da180-1e65-11eb-9ca9-dcbd056d22a0.png)

![image](https://user-images.githubusercontent.com/28277461/98092584-94170980-1e65-11eb-9c66-75c086e5f524.png)

![image](https://user-images.githubusercontent.com/28277461/98092644-a85b0680-1e65-11eb-88e3-d330a6ffaeb0.png)

Log file: (pretty much empty, but might as well include it)
[error.log](https://github.com/Innoxia/liliths-throne-public/files/5486786/error.log)

 - Game version: 0.3.9.9
 - OS: Windows 10
 - Using .jar or .exe: .jar (1.8.0_271)

AceXP:
This is caused by the latest Java version  (1.8.0_271), and reverting back to an earlier version should fix it.

LurkerHub:
Is that a known Java bug that will be fixed in a later version or a deliberate change that is here to stay?
If the latter, staying on an older version forever is not practical.

AceXP:
I think (but I am not certain) that it is related to this bug:

https://bugs.openjdk.java.net/browse/JDK-8256138

The tooltip does use a WebKit animation and if this is removed, the tooltips will show again (but less fancy).

But because it looks like this is a new issue that was caused by a recent change in Java, I think there is a very good chance that it will be fixed in later versions of Java.

LurkerHub:
This big is about Java 15,  the problem happens with Java 8

AceXP:
It says that 8u271 is also affected. Both by this reported bug and the change that is referred to.

AceXP:
PR #1449 was merged so this should work properly now. Could you check it and close this issue?

`Made the tooltip fade-in animation toggleable via the existing 'Fade-In' game option. This fixes the bug where tooltips were appearing blank when using some versions of java to run the game. (PR#1449 by CognitiveMist)`

PR #1449, opened 2020-12-06
Tooltip animation is toggleable

CognitiveMist:
Tooltip fade-in animation is now toggleable via the existing 'Fade-In' game option.

Addresses #1444:
When using Java 8u271, there is a bug with animations in JavaFX WebView (https://bugs.openjdk.java.net/browse/JDK-8256138). In LT, this causes tooltips to appear blank for users with that version of Java.

This change addresses that bug by restricting tooltip animation to the game's 'Fade-In' option, which is disabled by default. Without the animation, tooltips render normally.

No new graphical assets required. Tested on 0.3.9.9.

### DIFF
--- a/src/com/lilithsthrone/controller/MainController.java
+++ b/src/com/lilithsthrone/controller/MainController.java
@@ -2329,6 +2329,9 @@ public class MainController implements Initializable {
 	}
 	
 	public void setTooltipContent(String content) {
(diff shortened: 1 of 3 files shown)


lilalesbian:
Tried it out with the current dev version (0.3.10) and it works!

Innoxia:
Thank you for this!

PR #1453, opened 2020-12-20
Prevent both types of strap-on from being equipped together

AceXP:

Prevent a strap-on from being equipped when there is one already (in another slot).

Check if a strap-on is already equipped for items with the 'REQUIRES_NO_PENIS' tag.

Also use clothing slots to determine dildo colour, as they can be in different slots, not only PENIS.

No

Yes, tested with version 3.13

AceXP#0930


AceXP:
More witchcraft from @Innoxia has already solved the issue where a strap-on could not be equipped anymore. 

But it is still possible to equip a second one (in a different slot), this PR fixes that.

Innoxia:
Thank you for this!

PR #1454, opened 2020-12-20
minimumWingSizeForFlight should be 'THREE_LARGE' for feral quadrupeds

Stadler76:
  While a pegasus-morph requires the standard `WingSize.THREE_LARGE` to be able to fly, a feral pegasus needed `WingSize.FOUR_HUGE` which was inconsistent.

  Fixed the above by making `minimumWingSizeForFlight` depend on the body being feral or not instead of being hardcoded into the `LegConfiguration`.

  No.

  Yes, 0.3.10 Alpha

  `@Stadler#3007`


Innoxia:
Thank you for this!

PR #1457, opened 2020-12-22
fix bug with passive form of elementals

AceXP:

Fix a bug where having an elemental in a feral form leaves the left and right sidebar blank

Check if the feral subspecies of the elemental has it's SVGstrings initialised before accessing them

No

Yes, tested with version 3.11

AceXP#0930


Innoxia:
Thank you for this!

PR #1458, opened 2020-12-23
Sean challenge

### LINKED ISSUE #1434: Impossible challenge with Constable Hughes (Sean)
And again, sorry for my poor english

**Describe the bug**
There is only one chance to challenge Sean. If you change your mind, you can't challenge him again.

**To Reproduce**
1. Talk to Constable Hughes in Slaver Alley
2. Choose "Complain"
3. Then "Persist"
4. Then "Challenge"
5. Change your mind via "Back out"
6. Return to Constable Hughes and repeat steps 1-4
7. See that option "Challenge" does not work

**Additional information:**
 - Game version 0.3.9.9
 - OS: Windows 10
 - Using .exe

AceXP:

Resolves #1434 When challenged, Constable Hughes (Sean) will take of some clothes, but since these do not get replaced, backing out of the challenge and then challenging again will create errors.

Replace the displaced clothing after backing out (and after combat). Also, the text still mentions a cap and jacket, but these were replaced by a beret and stabproof vest some time ago. So this PR also updates the text to account for this.

No

Yes, tested on 3.11

AceXP#0930


Innoxia:
Thank you for this!

PR #1459, opened 2020-12-23
Pregnant male slimes

AceXP:

Prevent male (no vagina) slimes that are pregnant from changing back to flesh, as the birthing scenes only account for characters with a vagina.

Added a description for when the transformation is attempted. Also made some minor changes in the birthing scenes for slimes, since these also mention a vagina. And change the description of 'the place where your womb should be' to 'your belly' for pregnant male slimes.

No

Yes, tested with 3.11

AceXP#0930


Innoxia:
Thank you for this!

PR #1460, opened 2020-12-25
Cum Stud to give same Affection bonus from Milking assignment as Lactating fetish

Eliria:
Remade prev PR to also give Cum Stud holders the affection bonus granted to those with Lactating fetish. Since it was tested before and I've only copypasted the same OR conditional there should be no reason it wont work, but I have not yet compiled with verson 0.3.12

### DIFF
--- a/src/com/lilithsthrone/game/occupantManagement/slave/SlaveJob.java
+++ b/src/com/lilithsthrone/game/occupantManagement/slave/SlaveJob.java
@@ -370,7 +370,7 @@ public enum SlaveJob {
 		@Override
 		public float getAffectionGain(int hour, GameCharacter slave) {
 			float aff = this.affectionGain;
-			if(slave.hasFetish(Fetish.FETISH_LACTATION_SELF)) {
+			if(slave.hasFetish(Fetish.FETISH_LACTATION_SELF) || slave.hasFetish(Fetish.FETISH_CUM_STUD)) {
 				aff = 2f;
 			}
 			Cell c = this.getWorkDestinationCell(hour, slave);


AceXP:
I would think this affection gain for this job could be improved a little:

Only gain affection if a slave  gets milked and produces milk (from breast or crotchboobs) and has the lactation fetish.

Or if the slave has cum extracted and produces cum and has the cumstud OR masturbation fetish.

Or if the slave has girlcum extracted and produces girlcum and has the pussy slut OR masturbation fetish.

Eliria:
> Only gain affection if a slave gets milked and produces milk (from breast or crotchboobs) and has the lactation fetish.
> 
> Or if the slave has cum extracted and produces cum and has the cumstud OR masturbation fetish.
> 
> Or if the slave has girlcum extracted and produces girlcum and has the pussy slut OR masturbation fetish.

I did think about restricting it to actual usage but decided against, due to increasing complexity/specificity with low gameplay augmentation. The game is abstracting things at various aesthetic levels. While logically it makes more sense to do per-case it can be kept broad for game mechanic reasons... the affection gain can be thought of less as specifically the physical pleasure experienced on a case by case basis and more an overall experience of the job. Even if someone were assigned to assist others in getting milked, if they have the lactation / cum stud fetish they are still likely to find it an appealing job and thus gain affection from being in the pleasing task. I'm not sure of the exact aesthetics Inno is looking for so it's simply a mechanical expansion I'm doing right now.

If Inno were to request specific alterations I'd be confident in revising the PR's design changes. Otherwise I'll wait on feedback on the new change and incrementally propose retooling.

Innoxia:
I somehow managed to break the merging process on my end, so I've implemented this PR locally and pushed to the dev branch, crediting you for the change and this PR in the patch notes. Thank you for this!

PR #1467, opened 2020-12-29
Appearance fix for feral legConfigurations

Stadler76:
  Appearance for feral legs were hardcoded and didn't take various leg configurations into account. Fixed that.

  In addition to the fix mentioned above I switched from string concatenation to chain of appends in the chunks I updated.

  No.

  Yes, 0.3.13 Alpha

  `@Stadler#3007`


Innoxia:
Thank you for this!

PR #1471, opened 2021-01-03
Cache sub species

AceXP:

Improve performance by caching subspecies variables and only doing a recalculation when needed.

Save subspecies and return it instead of doing the calculation again and again. Reset subspecies when needed.
Removed a space in an assignment in Body.java that was not really necessary.

No

Yes, tested with version 0.3.13

AceXP#0930


Innoxia:
Thank you for this!

PR #1483, opened 2021-01-12
To flap or not to flap

Stadler76:
Ryus don't use wings to fly, even if they would have wings. Fixed that so that the wings aren't mentioned in such scenes, when the player can fly without wings.
I might have overlooked a few scenes where it is still mentioned, that the player uses his/her wings, but most scenes should now play correctly if the player can fly without wings.

In the `.java`-files I've modified I've used chain of appends instead of string concatenation where applicable.

No.

Yes, 0.3.13 Alpha

`@Stadler#3007`


Innoxia:
Thank you for this!

PR #1484, opened 2021-01-12
store result of getSubspeciesWeighting to improve performance

AceXP:

Improve performance by storing the result of getSubspeciesWeighting

Load race descriptions from characters/raceInfo if there is no bookEntries file (like the new Cougar subspecies)

Store the result of getSubspeciesWeighting and use this for the comparison and to assign it as a value

Only load from the bookEntries file if it exists, otherwise use the characters/raceInfo for parsing.

No

Yes, tested with 0.3.14

AceXP#0930


Innoxia:
Thank you for this!

PR #1486, opened 2021-01-14
Winged biped LegConfiguration for ferals

Stadler76:
This is a LegConfiguration only available for ferals, like bats or wyverns. Compared to avians they usually tend to walk on all fours by using their wings like forelegs.

I may have overlooked a few files where the current LegConfiguration is checked for. But basically everything, that works for bipeds works for winged-bipeds as well with a few exceptions with things, where normal arms and hands are needed for. For example `AbstraceLegType.handleLegConfigurationChanges()` doesn't consider `LegConfiguration.WINGED_BIPED` as this is not to be used in enchanting or any non-feral TF.

Basically no, as this is a feral only LegConfiguration, not to be used in enchanting. But for modders there *might* be the following assets needed:
- modifier_circle_tf_legConfig_wingedBiped
- statusEffects/race/raceBackgroundLegWingedBiped

Yes, 0.3.14 Alpha

`@Stadler#3007`


src/com/lilithsthrone/game/character/body/valueEnums/LegConfiguration.java
```diff
@@ -654,6 +654,65 @@ public void setLegsToDemon(GameCharacter character) {
...
+											InventorySlot.WEAPON_MAIN_2,
+											InventorySlot.WEAPON_MAIN_3,
+											InventorySlot.WEAPON_OFFHAND_1,
+											InventorySlot.WEAPON_OFFHAND_2,
+											InventorySlot.WEAPON_OFFHAND_3),
+									character.getLegType().getRace(),
+									"Due to the fact that [npc.nameHasFull] the feral body of [npc.a_legRace], [npc.she] cannot wield regular weapons!",
+									Util.newArrayListOfValues(
+											ItemTag.FITS_ARM_WINGS,
+											ItemTag.FITS_ARM_WINGS_EXCLUSIVE)));
```
Stadler76, line 702: As they have no wings or paws or similar to 'hold' any weapon, only weapons that are designed to be attached to wings should work here. e. g. some kind of spikes or blades since a strike from arm wings is more, like a [wing buffet](https://classic.wowhead.com/spell=18500/wing-buffet).
The same could be applied to feral avians, but this might be debatable since they don't work well for a wing buffet.
And since not every avian body has wings instead of arms you might wanna use something, like:
```java
if(character.isFeral()) {
	return Util.newArrayListOfValues(
		!character.getArmType().allowsFlight() ? null : new BodyPartClothingBlock(
		// [snipp]
		Util.newArrayListOfValues(
			ItemTag.FITS_ARM_WINGS,
			ItemTag.FITS_ARM_WINGS_EXCLUSIVE)));
}
```

Stadler76:
Here's a before (left) and after (right) to show the main reason for this PR:

Innoxia:
Thank you for this!

PR #1488, opened 2021-01-16
Standardize all stroke colors to RGB black instead of CMYK.

DSGMods:
Fix graphical issues that arise when selecting very dark dye colors on items due to some stroke colors being set to CMYK black (#231f20) instead of RGB black (#000000)

All instances of the stroke color #231f20 replaced by #000000

All modified assets are included

Tested on 0.3.14

dsg#4931


Innoxia:
Thank you for this!

PR #1489, opened 2021-01-16
Implement Affinity enum and affinity attribute for feral names in race.xml

Stadler76:
This is mainly to allow separate naming for aquatic subspecies that aren't mer-tailed.

I've added the container class `LegConfigurationAquatic` and added to the package `com.lilithsthrone.game.character.body`.
This is simply to have a class, that combines the `LegConfiguration` and the boolean aquatic since every entry has two attributes.
See the changes to [res/race/dsg/dragon/race.xml](https://github.com/Innoxia/liliths-throne-public/commit/95d82e726d1904b00536990d965be6541e40d1c1#diff-e895b15a9cc0c9545a978d227b72e6804b52946133c7ca0fdf993b7422800fbb) for an example, how this works.

**Update (30.10.2022):**
I've refactored `LegConfigurationAquatic` to `LegConfigurationAffinity` and added the `Affinity`-enum.
For example `<aquatic>true</aquatic>` is replaced with `<affinity>AQUATIC</affinity>` and compared to `<aquatic>`, `<affinity>` is optional and defaults to `AMPHIBIOUS` as seen in [res/mods/innoxia/race/hyena/subspecies/spotted.xml](https://github.com/Innoxia/liliths-throne-public/pull/1489/commits/584d37bd6673e276a4e507f377cff3c11fdc7488#diff-307e36d1570838bee322a290cadb1c36dfa8fe7e4a097babf324429a2075302f).
I've waded through all subspecies.xml-files and either removed `<aquatic>false</aquatic>` or replaced it with the `<affinity>`-tag, depending on which modders left the comments intact (See the `spotted.xml` linked above).

No

Yes, 0.4.6.6 Alpha

`@Stadler#3007`


Innoxia:
Thank you for this!

PR #1490, opened 2021-01-16
Enforcer Item Update

DSGMods:
Modernize and update the XMLs of all Enforcer items.

General changes:
Deleted all comments as these were often several generations old and potentially misleading to new modders or tinkerers.
Fixed miscellaneous typos and incorrect parser tags.

Weapons:
Changed all weapons that previously applied status effects through hitText to use the new effect elements instead.
Changed all status effects to be applied on critical hits instead of on every hit or through random chance.

Added special attacks to the Pepperball Rifle and Pistol.
 - The Mag Dump attack for the Pepperball Rifle can hit up to 15 times its base damage before modifiers, multiplier is determined at random.
 - The Seven Rounds Rapid attack for the Pepperball Pistol can hit up to 7 times for its base damage before modifiers, multiplier is determined at random.

Split the liquid stun gun's silly mode function into its own item that can be spawned via the debug menu.
 - The new, discrete, LSG-9000 coom gun item will now make the target ingest the attacker's fluids on critical hit under certain conditions.
 - Attackers with more than 150 ml of milk can now make the the target drink it if the target's mouth is not covered.
 - Attackers with more than 150 ml of cum have a 50% chance of either making the target drink it if the target's mouth is not covered, or hitting the target's groin if the target's groin is not covered.
 - If the target has a vagina without a hymen when hit in the groin, the attacker's cum will be ingested with all the usual consequences.

Clothing:
Added a test Enforcer beret with no sticker requirements to the debug menu.
Updated Enforcer beret with some description text that was lost during sticker work.

All new graphical assets are included. Please note that PR#1488 should be merged first.

Tested on 0.3.14

dsg#4931


Innoxia:
Thank you for this!

PR #1493, opened 2021-01-23
Use only a single factory for XML work

AceXP:

Improve performance and memory usage.

Use a single instance of the TransformerFactory, DocumentBuilderFactory and DocumentBuilder. This means that there is less overhead in many cases having to do with XML files, like reading files, saving files. Also memory usage is improved.

Note: I also added some remarks to the to_do list, and removed a few items that were done already.

No

Yes, tested with different saves with version 0.3.14

AceXP#0930


Innoxia:
Thank you for this!

PR #1494, opened 2021-01-24
Change antennae counting for tooltip

debouchere:
Change to account for less/more than two antenna per row.

Change tooltip for antenna to reflect exact number of antenna

Change how tooltip for antenna counts from `2 * rows` to `number per row * rows`

No

Yes, 0.3.14 alpha

deboucher#8190


Innoxia:
Thank you for this!

PR #1498, opened 2021-02-03
More code pretty for race modding

Stadler76:
This is the followup-PR for more code-pretty and simplifications for race modding.

New methods added:
- `Body.java`:
  - `hasWings()`: Moved from `GameCharacter` to `Body`
  - `getVaginaType()` and `setVaginaType()`
  - `isFaceHuman()`: For better support of partials. See `res/race/dsg/dragon/subspecies/coatl.xml` for a usage example.
  - `randomTypeFrom()` Selects a random 'Type' from the list of `BodyPartType`. See `res/race/dsg/dragon/subspecies/dragon.xml` for an example.
- `BodyPartTypeInterface.java`:
  - `isOneOf(...)`. Checks, if the BodyPart is among the list. See `res/race/dsg/dragon/subspecies/dragon.xml` for an example.

No.

Yes, 0.3.15 Alpha

`@Stadler#3007`


src/com/lilithsthrone/game/character/body/types/BodyPartTypeInterface.java
```diff
@@ -131,4 +132,15 @@ public default String getTransformName() {
...
 	}
+
+	/**
+	 * Checks, if the given BodyPartType is in the list of BodyPartTypes
+	 * @param values The list of BodyPartTypes to match the type against
+	 * @return true, if the type is among the list.
+	 */
+	public default boolean isOneOf(BodyPartTypeInterface... values) {
+		ArrayList<? extends BodyPartTypeInterface> list = new ArrayList<>(Arrays.asList(values));
+		list.removeIf(e -> e==null);
```
CognitiveMist, line 143: I don't think removing nulls is necessary, as `contains()` shouldn't be tripped up by it, and this results in walking the list an extra time (even if it is always a very short list). If this is indeed the case, I think the entire function would decompose to just the one-liner `Arrays.asList(values).contains(this)`. (The creation of a new ArrayList is unnecessary here, since the list is not modified.)

If I am wrong, then the first line could use `Util.newArrayListOfValues`, but that change would be of lesser importance.

src/com/lilithsthrone/game/character/body/Body.java
```diff
@@ -6386,4 +6413,7 @@ public void setTakesAfterMother(boolean takesAfterMother) {
 		this.takesAfterMother = takesAfterMother;
 	}
 
+	public BodyPartTypeInterface randomTypeFrom(BodyPartTypeInterface... values) {
```
CognitiveMist, line 6416: If you feel inclined to write a doc comment for this, it might be good to mention that simple weighted averages can be achieved by listing a more desireable outcome more than once.

Stadler76: I intend to add this after a quick test:
```java
	/**
	 * Returns a randomly chosen BodyPart-type from the list of types (e. g. multiple WingTypes)
	 * provided as parameters. To give one or more types more weight these types can be repeated.
	 * @param values List of BodyPartTypes to choose from randomly
	 * @return The randomly chosen type of the corresponding BodyPart
	 */
 ```

src/com/lilithsthrone/game/character/body/types/BodyPartTypeInterface.java
Stadler76: Well, The case `Arrays.asList(null).contains(null)` would theorhetically return true but that can't happen, since `this` can't be `null` without causing an NPE.

CognitiveMist (approved):

Innoxia:
Thank you for this!

PR #1503, opened 2021-02-09
secondaryColour and tertiaryColour default to colour in subspecies.xml

Stadler76:
In subspecies XMLs the `<secondaryColour>` and `<tertiaryColour>` should default to `<colour>`

If `<secondaryColour>` or `<tertiaryColour>` are empty they default to `this.colour` now.

No.

Yes, 0.3.15 Alpha

`@Stadler#3007`


Innoxia:
Thank you for this!

PR #1506, opened 2021-02-21
Self tf menu

AceXP:

Fix descriptions when using the transform menu for races that can self-transform and aren't slimes or demons.
Added hidden perks to keep track of the number of tails for Youkos (max 9). Change the options in the self TF menu for Youkos. They can switch between Youko, fox and human.

Enable changing secondary coverings colours for Elementals, and changed ice description to `glittering` instead of `shimmering`.

Fix the error where in Helena's slave creator, non-human coverings like fur won't appear.

Add a isSelfTFMenu() check, if true get mostly the same options as a demon tf, but with `your innate powers` instead of `your demonic powers`.
For Youkos remove several options (arm count, cloaca options, horns) in the TF menu. The number of tails is limited and can only be increased by a special method (like the Inno-Chan's gift). A new getMaxTailCount() method is used to determine a Youko's stats. 

No

Yes, tested with 0.3.18

AceXP#0930

Perk for Youko:
YoukoPerk

Youko tails:
ChangeYoukoTail


Innoxia:
Thank you for this!

PR #1510, opened 2021-02-21
Fix arachnid LegConfiguration removing the tail.

Stadler76:
Changing your LegConfiguration to 'arachnid' removed the tail aka abdomen.

- Added `Tail.class` to `getFeralParts` for `LegConfiguration.ARACHNID`
- Changed `isTailLostOnInitialTF` to false for `LegConfiguration.ARACHNID`
- Included `<type>charisma_spider_tailFluffy</type>` in `<tailTypes>` for spooder `racialBody.xml`

No.

Yes, 0.3.15 Alpha

`@Stadler#3007`

### DIFF
--- a/res/race/charisma/spider/racialBody.xml
+++ b/res/race/charisma/spider/racialBody.xml
@@ -108,6 +108,7 @@
 	<cumProduction>80</cumProduction>
 	<tailTypes>
 		<type>charisma_spider_tail</type>
+		<type>charisma_spider_tailFluffy</type>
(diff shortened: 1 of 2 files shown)


Innoxia:
Thank you for this, but I fixed the spinneret issue in the end by making the AbstractLegType 'spinneret' boolean apply only when the character's leg configuration is ARACHNID. As such, I was able to set spider leg types as spinneret=true.

PR #1516, opened 2021-03-13
Unenchantable item tag

cimds:
Adds a new item tag called UNENCHANTABLE, which greys out an item's enchant button in the inventory as if it were a normally unenchantable item type.

The tag works by modifying the nine if statements in InventoryDialogue that check if an item is an unenchantable item type to also check if they have this item tag.

The purpose of this tag is to potentially allow unique and gimmick items to retain their identity without a player stripping them of all enchantments in order to gain a set bonus with no downside, and to allow an item to increase enchantment capacity to make its enchantments free without worrying about the player removing them to gain enchantment capacity instead. The reason this tag was created was to take advantage of the fact that enchantments on an item can vary, while set bonuses are fixed to a specific effect. The idea is that an item could have a special move that permanently modifies its enchantments, which would be undermined by the player being able to enchant the item.

### DIFF
--- a/src/com/lilithsthrone/game/dialogue/utils/InventoryDialogue.java
+++ b/src/com/lilithsthrone/game/dialogue/utils/InventoryDialogue.java
@@ -29,6 +29,7 @@ import com.lilithsthrone.game.dialogue.responses.ResponseEffectsOnly;
 import com.lilithsthrone.game.dialogue.story.CharacterCreation;
 import com.lilithsthrone.game.inventory.ColourReplacement;
 import com.lilithsthrone.game.inventory.InventorySlot;
+import com.lilithsthrone.game.inventory.ItemTag;
 import com.lilithsthrone.game.inventory.Rarity;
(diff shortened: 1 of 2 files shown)


src/com/lilithsthrone/game/dialogue/utils/InventoryDialogue.java
```diff
@@ -1113,9 +1113,9 @@ public void effects(){
 								}
 								
 							} else if(index == 5) {
-								if(item.getEnchantmentItemType(null)==null) {
+								if(item.getEnchantmentItemType(null)==null || item.getItemTags().contains(com.lilithsthrone.game.inventory.ItemTag.UNENCHANTABLE)) {
```
AceXP: If you add _import com.lilithsthrone.game.inventory.ItemTag;_ (after import _com.lilithsthrone.game.inventory.InventorySlot;_ for example), you can shorten this to _item.getItemTags().contains(ItemTag.UNENCHANTABLE)_ which makes it easier to read.

PR #1517, opened 2021-03-15
Unenchantable item tag

cimds:
Adds a new item tag called UNENCHANTABLE, which greys out an item's enchant button in the inventory as if it were a normally unenchantable item type.

The tag works by modifying the nine if statements in InventoryDialogue that check if an item is an unenchantable item type to also check if they have this item tag.

The purpose of this tag is to potentially allow unique and gimmick items to retain their identity without a player stripping them of all enchantments in order to gain a set bonus with no downside, and to allow an item to increase enchantment capacity to make its enchantments free without worrying about the player removing them to gain enchantment capacity instead. The reason this tag was created was to take advantage of the fact that enchantments on an item can vary, while set bonuses are fixed to a specific effect. The idea is that an item could have a special move that permanently modifies its enchantments, which would be undermined by the player being able to enchant the item.

(This pull request briefly closed and had to be resubmitted. This is because I'm still learning how to use git and was trying to restructure my fork so it's in a more usable state on my end, so I would more easily be able to make more PRs than just this one. Sorry.


Innoxia:
Thank you for this!

PR #1530, opened 2021-04-05
Talon slash

AceXP:

Add a special attack to characters that have taloned feet (i.e. harpies). The base damage is 20 (same as a kerambit). If the attack does damage, extra damage is applied as a critical effect.

Add a special attack TALON_SLASH. Updated the constructor of the AbstractCombatMove to allow the equipWeighting to be passed in. Updated the display of the move to not show the equip/unequip line when in combat.

A new icon is provided (with help from the developers chat).

Yes, tested with 0.3.21

AceXP#0930
Attack information:
TalonSlash

Attack effect:
TalonsEffect


Innoxia:
Thank you for this!

PR #1531, opened 2021-04-05
Spawn rates

AceXP:

Show the chances of encountering various races in the current location.

Added an option in the debug menu (misc tab, 2nd page) to show the spawn rates in the current location. This will show the overall, masculine and feminine rates for each race. These are the values from the weighted getWorldSpecies map shown as a percentage of the total, based on spawn frequency and spawn preferences for the location, race and gender. The actual races encountered may be different, as that depends on how encounters are generated (but most use the same weighted getWorldSpecies map).

No

Yes, tested with 0.3.21

AceXP#0930

Spawn rates in one of Dominion's alleys:
AlleySpawn

Spawn rates in the Foloi Forests (not implemented as encounters when this PR was created):
FoloiForestSpawn


Innoxia:
Thank you for this!

PR #1537, opened 2021-04-13
Removed the need of the toc.xml

Stadler76:
Deprecated the tox.xml in favor of file name based sorting.

- Fixed the issue with wrong case in the folder name: `patchnotes` → `patchNotes`
- Deprecated and removed the `toc.xml` in favor of sorting based on the file name:
  - Every file is prefixed with the release date in YYYYMMDD-format. e. g.: `20210413-patchnotes-0.3.22.html`
  - Filenames are matched against the RegEx-pattern: `^\d{8}-(preface|patchnotes)-` case insensitively.
    So if you don't want patchnotes to be displayed but don't want to delete it yet I suggest renaming it to `20210413-old-patchnotes-0.3.22.html` for example.
  - Only the newest preface-file is loaded, while the others are skipped, so it's not mandatory to delete or rename old `preface.html`s
  - The date is mainly used for sorting, so it doesn't have to be exact. e. g., if you have multiple releases on the same day you could simply 'cheat' with the dates. For example by incrementing them.
- The patch notes are now loaded just in time. Meaning: `private static void loadPatchNotes` is now `public static String getPatchNotes` and at the same time `Main.patchnotes` is private.

No.

Yes, 0.3.21 Alpha

`@Stadler#3007`


Innoxia:
Thank you for this!

PR #1541, opened 2021-04-27
Stops keys for locked clothing being removed during sex to avoid key loss on automatic re-equip

debouchere:
Keys for locked items unequipped during sex are currently removed, even though the item is automatically re-equipped after sex.  This stops the key from being removed from the key map for sealed items unequipped during sex.

- See above

No

Checked on 0.3.21 Alpha, didn't seem to duplicate keys or do anything weird.

- deboucher#8190


Innoxia:
Thank you for this!

PR #1550, opened 2021-05-15
Typos and consistency

AceXP:
- What is the purpose of the pull request?

Fix several typos, consistency errors and (minor) bugs.

- Give a brief description of what you changed or added.

Typos:
Dragon leg transformation mentions fur, should be scales.
Mommy's nice and full and today -> Mommy's nice and full today.
Remove extra ) at the end of dialogue with pregnant npc.
out of this area ... -> out of this area...
candi pouts -> Candi pouts.
greedily kissing and kicking the cougar-girl -> greedily kissing and licking the cougar-girl.
take out the shotgun -> take out the rifle (DSG HLF bolt rifle)
centuar -> centaur

Consistency:
River Huber -> River Hubur.
One Night in June -> One Day in June.
Bat cavern attacker mentions two-hundred and fifty flames, should be five hundred flames.
Red light district -> Red-light district.
Nyan: I'm on the second floor -> I'm on the third floor.
Opening hours of the Watering Hole should display in 12/24 hour clock format

Bugs:
Parser error with kilometers/miles when the world map is mentioned.
Parser error in the stocks if you have a companion.
appears to be in his past seventy -> appears to be past seventy
Several text fixes and other issues in Helena's quest and Dominion Express content
Fixed switched text for single/multiple tails while petting
Thoughts should not be muted (also change dialogue for alleyway demon attacker)
Minor octopus morphs spawned with eight HUMAN legs.
Changed some occurences of step and walk to user parser commands [pc.step] etc.
After freeing a slave a Freedom certificate should be removed from the inventory.
Clothing that is discarded on use (condom, web seals) should be removed from the game when doing Unequip all.
Fixed info text for changing pubic hair mentioning no genitalia when that is not true.
Fixed (not solo) handjob climax looking weird when the character is giving themselves a handjob.
Fixed elemental birth date of 29 february leading to Exception
Fixed self-TF menu leading to NPE with `<ableToSelfTransform>true</ableToSelfTransform>` race
Fix issues with Bunny/Loppy pregnancy (also move Loppy images to KT80 folder)

New features:
Added missing penetration length code for tentacles.
Add the rolled-up posters to the encyclopedia when you get them, as you can never see them otherwise.
Add crawl/slide as movement verbs for cephalopods.

- Are any new graphical assets required?

No

- Has this change been tested? If so, mention the version number that the test was based on.

Yes, tested with 0.4.0

- So we have a better idea of who you are, what is your Discord Handle?

AceXP#0930

PR #1556, opened 2021-06-04
Fix typo in ELSE statement

Jhortrax:
Fixed the capitalization of the ELSE statement in /res/txt/misc/lyssiethReveal.xml#L1063.
Without the capitalization, the unit description is rendered as `kilometeres#ELSemiles` if the Metric Sizes option is enabled, and is omitted if not.

2021-06-04T02-08-06

No new graphical assets required.

This change has not been tested.

Discord handle `DaniDipp#0001`.


AceXP:
This error (and about 30 others) is also fixed in PR #1550.

Jhortrax:
oh, my bad. I wish GitHub would have shown me that PR when I was looking at the file history.

PR #1559, opened 2021-06-10
Offspring seed

AceXP:

Improve performance and reduce save file size.

No

Yes, tested with the latest 0.4.0.8 version

AceXP #0930


AceXP:
This PR is now combined with #1561 Save folders into the single PR #1563.

AceXP:
Closing this one, as now included in PR #1563

PR #1560, opened 2021-06-10
Guest jobs

Maxis010:
Allows guests to take on some of the more mundane slave jobs

Changed UI to allow assigning work to guests that don't have jobs
Changed OccupancyUtil to support guests working slave jobs

No

Functional testing on 0.4.0.9, additional testing recommended
Dialog needs correcting as guests are being refered to as slaves, using their room dialog when interacting from their work tile & at work dialog on the guest room isn't detecting their hours correctly

Maxis010#3060


Maxis010:
Haven't had a chance to look at the various issues I reported on discord but they are not quick fixes

Innoxia:
Thank you for this! 💼

PR #1561, opened 2021-06-13
Save folders

Maxis010:
Converts save files into save folders, vastly reduces save time by only saving NPCs that have changed since the last save

Changed all save & load methods to support loading both formats but only save to folder format

No

Yes, although I would recommend more testing and STRONGLY recommend users back up their saves when this gets added

Maxis010#3060

### DIFF
--- a/src/com/lilithsthrone/main/Main.java
+++ b/src/com/lilithsthrone/main/Main.java
@@ -879,8 +879,7 @@ public class Main extends Application {
 			File[] childFiles=file.listFiles();
 			if(childFiles!=null) {
 				for(File child : childFiles) {
-					boolean result =child.delete();
-					System.out.println(result);
+					child.delete();
 				}
 			}
 			file.delete();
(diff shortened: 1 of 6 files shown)


src/com/lilithsthrone/main/Main.java
```diff
@@ -796,25 +796,15 @@ public static String getQuickSaveName() {
 		if(!Main.game.isStarted()) {
 			return "QuickSave_intro";
 		}
-		return "QuickSave_"+Main.game.getPlayer().getName(false);
+		return "QuickSave_"+Main.game.getPlayer().getSurname();
```
AceXP, line 799: Can surname be empty for the player? If so, need an additional check for that.

Maxis010: Surname can be an empty string, at which point this would become Quicksave_

AceXP:
This PR is now combined with #1559 Offspring Seed into the single PR #1563.

Maxis010:
Closing PR as I have added further updates to #1563 with AceXP

PR #1563, opened 2021-06-19
Offspring seed saves

AceXP:

Combine PR #1559 OffspringSeed with PR #1561 Save folders

When loading an old save, any existing NPCOffspring NPC's that have not been encountered yet are converted to the new offspringSeed class. This is a one-time only event. It would be best to make a new save straight away. But also see the advise below.

Also some minor tweaks for dialogue when offspring is not birthed by the player, but by an incubator.

Converts save files into save folders, NPC's and offspringSeed are each saved in their own file. Since offspringSeed only needs to be saved once (they will never change) this reduces save time.

No

Yes, tested with 0.4.0.10 (and earlier versions). Although the recommendation is to do more testing and STRONGLY recommend users to back up their saves when this gets added. There are multiple changes in the way saves work with these changes.

AceXP #0930 and Maxis010 #3060


src/com/lilithsthrone/game/Game.java
```diff
@@ -518,22 +513,17 @@ public static void exportGame(String exportFileName, boolean allowOverwrite) {
...
+		if(dir.exists()) {
+			if(!allowOverwrite) {
+				Main.game.flashMessage(PresetColour.GENERIC_BAD, "Save already exists!");
+				return;
+			} else {
+				overwrite = true;
 			}
 		}
-
+		dir.mkdir();
```
Maxis010: Looks like this slipped through when I was experimenting with per character folders

src/com/lilithsthrone/game/Game.java
```diff
@@ -655,46 +645,75 @@ public static void exportGame(String exportFileName, boolean allowOverwrite) {
...
 		}
-
-
+		
+		// Add all offspringSeed:
+		try {
+			String id = "";
+			for(OffspringSeed offspringSeed : Main.game.getOffspringSeedMap().values()) {
+				id = offspringSeed.getId().replace(",", "_");
+				if(!new File("data/saves/"+exportSaveName+"/"+id+".xml").exists()) {
```
Maxis010, line 700: I assume offspringSeed won't change after creation so it's safe to only have an exists() check?

src/com/lilithsthrone/game/Game.java
AceXP: This probably stems from the merge and needs to be looked into.

src/com/lilithsthrone/game/Game.java
AceXP: Yes

PR #1567, opened 2021-06-19
Don't parse the patch notes.

Stadler76:
The patch notes were run through the parser which could cause NPEs and the patch notes not displaying.

Added and implemented
```java
/**
 * @return Whether to content of the dialogue should run through the parser. Will almost always be {@code true}.
 */
public boolean isContentParsed() {
	return true;
}
```
in `DialogueNode.java`
and tidied the HTML in `res/patchNotes/20210607-patchnotes-0.4.0.8.html`.

No.

Yes, 0.4.0.10 Alpha

`@Stadler#3007`


Innoxia:
Thank you for this!

PR #1568, opened 2021-06-20
Replace hasSexCountWith method with getTotalTimesHadSex

AceXP:

Refactoring: Remove the hasSexCountWith() method and replace all it's uses with the getTotalTimesHadSex() method.

The hasSexCountWith method returns true when there is an entry for the npc in the sex count, even if all counts are zeroes. This leads to inconsistencies in dialogues. Only a single method should be used.

No

Yes.

AceXP #0930


Innoxia:
Thank you for this!

PR #1569, opened 2021-06-20
Outfit regions

AceXP:

Introduce a new setting for outfits which makes it possible to apply an outfit to an entire region (i.e. collection of maps).

Added a new property worldRegions to outfits. An outfit can be applied if either the region or the worldtype applies.

Also changed the sequence in which various properties are checked in the isAvailableForCharacter, checking the most significant properties first.

No

Yes, tested with 0.4.0.10 (and earlier)

AceXP #0930


Innoxia:
Thank you for this!

PR #1570, opened 2021-06-20
Slave for sale

AceXP:

Introduce a new dedicated class for slaves that can be bought in Slaver Alley.

Previously, these slaves were spawned from the DominionAlleywayAttacker class. But that class has logic to determine the new NPC based on player location, so this will produce strange/unwanted results when the player is not in Dominion/Submission.
The new class only generates NPC that can spawn naturally in Dominion or Submission, or the SLAVER_ALLEY_SLAVERY_ADMINISTRATION location.

No

Yes, tested with 0.4.0.10 (and earlier)

AceXP #0930


Innoxia:
Thank you for this!

PR #1571, opened 2021-06-22
Proposal: changing HashMaps/HashSets to EnumMaps/EnumSets

WhatsInTheCrux:

It appears that somewhat randomly EnumMaps or EnumSets are (correctly) used when the maps and sets are only using the enum values as keys. But a lot of these opportunities are still using the regular collections. I considered replacing them but before I go all in wanted to make a pull request with the proposed idea before unnecessarily putting time into something that is undesirable. I have tested this and have not run into any errors trying to approach every property in some way ingame.

One file just for the proposal purpose: GameCharacter. Since GameCharacter is being instantiated a lot for every NPC/Slave/Encounter it will also be using these properties a lot. The diff should speak for itself: I changed HashMaps to EnumMaps and HashSets to EnumSets where applicable, that is: the HashMaps and HashSets are using Enum values for keys and can therefore be replaced with their optimized counterparts.

None.

This is tested on the dev preview 0.4.0.10 or more precise, its previous commit 604c2259fb08c6d46721d8d84eba89602c066c27

Crux#9106

@Innoxia (because I think the answer can be as simple as "seems promising" or "heck no").

Thanks for the great gaming nonetheless.

### DIFF
--- a/src/com/lilithsthrone/game/character/GameCharacter.java
+++ b/src/com/lilithsthrone/game/character/GameCharacter.java
@@ -329,14 +329,14 @@ public abstract class GameCharacter implements XMLSaving {
 	protected int ageAppearanceDifference;
 	
 	protected Occupation occupation;
-	protected Set<Occupation> desiredJobs;
-	protected Set<PersonalityTrait> personalityTraits;
+	protected EnumSet<Occupation> desiredJobs;
+	protected EnumSet<PersonalityTrait> personalityTraits;
 	protected SexualOrientation sexualOrientation;
 	private float obedience;
 
 	private int experience;
 	private int perkPoints;
-	private Map<PerkCategory, Integer> perkCategoryPoints;
+	private EnumMap<PerkCategory, Integer> perkCategoryPoints;
 
 	protected List<Artwork> artworkList;
 	private int artworkIndex = -1;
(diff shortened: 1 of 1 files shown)


src/com/lilithsthrone/game/character/GameCharacter.java
```diff
@@ -422,10 +422,10 @@
 	protected DialogueNode enslavementDialogue;
 	protected AbstractClothing enslavementClothing;
 	
-	protected Map<SlavePermission, Set<SlavePermissionSetting>> slavePermissionSettings;
+	protected EnumMap<SlavePermission, EnumSet<SlavePermissionSetting>> slavePermissionSettings;
 	
 	protected SlaveJob[] workHours;
-	protected Map<SlaveJob, Set<SlaveJobSetting>> slaveJobSettings;
+	protected EnumMap<SlaveJob, EnumSet<SlaveJobSetting>> slaveJobSettings;
```
CognitiveMist, line 428: This in an example of an unnecessary change.

src/com/lilithsthrone/game/character/GameCharacter.java
```diff
@@ -554,26 +554,35 @@ protected GameCharacter(
 		enslavementDialogue = null;
 
 		workHours = new SlaveJob[24];
-		for(int i=0; i<workHours.length; i++) {
-			workHours[i] = SlaveJob.IDLE;
-		}
+        Arrays.fill(workHours, SlaveJob.IDLE);
 		
-		slaveJobSettings = new HashMap<>();
+		slaveJobSettings = new EnumMap<>(SlaveJob.class);
```
CognitiveMist, line 559: This is an example of a necessary change.

CognitiveMist (commented):
I can't comment on the utility of this change since I haven't done much research on the topic, but there are more changes here than necessary.

The new `EnumMap`/`EnumSet` keywords only need to appear when creating new objects.

Essentially, type declarations, whether declaring variables or providing a function's signature, should probably remain as `Map`/`Set`, so as to maintain the black box, and reduce coupling.

Hopefully that's clear, but feel free to ask if you're unsure.

PR #1572, opened 2021-06-23
Duplicate entry for antenna, breast is a required final type.

WhatsInTheCrux:
Small fix


Innoxia:
Sorry that I missed this; I've already fixed this on my end, so there's no need for this to be merged now. Thank you anyway!

PR #1573, opened 2021-06-25
Removed unnecessary copy when accessing the unique item count of an inventory.

AtelierAmber:

Removal of an unnecessary copy during inventory accesses.

Added a function to get the unique item count from an AbstractInventory.  When clicking on an item in the inventory, the function `com.lilithsthrone.game.inventory.CharacterInventory;getUniqueItemCount()` is called 108 interspersed between rendering updates and inventory accesses. The current version uses the function `com.lilithsthrone.game.inventory.CharacterInventory;getAllItemsInInventory()` and gets the size of the result. The issue is that `Collections.unmodifiableMap(...)` copies the entire contents of the input map into a new map. This results in an excessive cleanup job for the garbage collector, and slows the entire game down immensely, not to mention using the inventory is a laggy experience from the get go. By removing all 108 unnecessary copies the inventory screen works very smooth and the entire game has significantly less lag caused by an excessive increase in memory usage. 

No

Yes, latest as of 7-12-2021, version 4.1, commit [81a1fc0](https://github.com/Innoxia/liliths-throne-public/commit/81a1fc0b147c557baf890e25224da14fb2f487c1)

Amber💞#5285

### DIFF
--- a/src/com/lilithsthrone/game/inventory/AbstractInventory.java
+++ b/src/com/lilithsthrone/game/inventory/AbstractInventory.java
@@ -72,6 +72,10 @@ class AbstractInventory<T extends AbstractCoreItem, U extends AbstractCoreType> 
 		return Collections.unmodifiableMap(duplicateCounts);
 	}
 
+	int getUniqueItemCount(){
+		return duplicateCounts.size();
+	}
+
 	int getQuestEntryCount() {
 		return (int) duplicateCounts.keySet().stream().filter(e -> e.getRarity().equals(Rarity.QUEST)).count();
 	}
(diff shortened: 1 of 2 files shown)


AtelierAmber:
Updated to 0.4.1 public release

AceXP:
It looks like the delta from your commit is not quite right, as it lists 137 files (from Innoxia's latest commit). So you might want to look into this and just create the clean commit from your changes.

AtelierAmber:
> It looks like the delta from your commit is not quite right, as it lists 137 files (from Innoxia's latest commit). So you might want to look into this and just create the clean commit from your changes.

Went ahead and rebased onto dev, so it should be fine now. Never had that happen in other pull requests, I might have screwed something up somewhere.

Innoxia:
Thank you for this!

PR #1577, opened 2021-07-10
Wet/Dry Rework

Maxis010:
Standardise Save/Load XML methods
Java keyword optimisations
Rewrite existing code to be easier to read/modify

XMLSaving and all classes that impliment it
Added new Data static to standardise save process & directory names (in case something is changed later)

No

Currently can not be built

Maxis010#3060


Maxis010:
Closing this as I have no timeline on completion and there are other smaller reworks that will interfere

PR #1578, opened 2021-07-11
Check hidden status for prepare and deny

AceXP:

Don't allow NPC's to react to the MC orgasming when the MC is hidden.

Added checks for the prepare and deny actions to see if the target is hidden, and also if the target is really the character orgasming.

No

Yes, tested with 0.4.1

AceXP#0930


Innoxia:
Thank you for this!

PR #1581, opened 2021-07-23
Pet name expansion

CognitiveMist:
Change the pet name logic to be based on a list, instead of
hardcoded. Also add a bunch of new adaptive pet name options.

Tested on 0.4.1.

### DIFF
--- a/src/com/lilithsthrone/game/character/GameCharacter.java
+++ b/src/com/lilithsthrone/game/character/GameCharacter.java
@@ -4005,22 +4005,54 @@ public abstract class GameCharacter implements XMLSaving {
 		return petNameMap;
 	}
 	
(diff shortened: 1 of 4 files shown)


CognitiveMist:
I also had the idea of defining pet names in the map as triplets, and then just ripping out the adaptive pet name code completely.

Then, the UI could have two slots for defining the feminine and masculine names (maybe collapsing to a single box depending on a checkbox). This would, for example, allow a masculine character to be called "Mistress", if the player so desired, in addition to letting the player define the adaptive pairs more freely.

This alternate change would be a larger PR (and also replace this one), but I'd be happy to attempt an implementation if there is interest.

CognitiveMist:
This PR has been superseded by #1591.

PR #1582, opened 2021-07-23
Enhancement: NPCs can spawn dirty

CognitiveMist:
If an NPC has the CUM_ADDICT fetish, or they are an arcane storm
attacker, there is a 10% chance that they will spawn with some
dirty inventory slots. If chosen, the NPC will have between 1 slot
and half their slots marked as dirty.

No assets required, does not affect save data format.

Tested on 0.4.1.


Innoxia:
Thank you for this!

PR #1583, opened 2021-07-23
QoL: Alter corruptive action icon when bypassed by fetish

CognitiveMist:
If a corruptive action is bypassed by fetish, it will now be
displayed as greyscale, instead of full-color like a normal
corruptive action, to signify that no corruption will be gained.

Tested on 0.4.1.


Innoxia:
Thank you for this!

PR #1584, opened 2021-07-23
QoL: Buy-max when unable to afford buy-all

CognitiveMist:
When the player does not have enough flames to use "Buy (All)",
but has enough to buy at least one item, the game will offer
"Buy (Max NN)" in the "Buy (All)" slot, where NN is the amount
that will be purchased.

Tested on 0.4.1.


Innoxia:
Thank you for this!

PR #1590, opened 2021-08-09
Fixes the Infinite recursion on loading player saves.

Jtecx:
Fixes Infinite recursion on loading player save if player has slaves/followers.

 Technically, I didn't change anything. Cognitive Mist did. But I found the problem and put his fix in and tested, it working so...
 Fixes infinite recursion from chars when loading an existing save, as well as a new save.

 Nope.

 Yes, 0.4.1.3

 @Jtecx

👍


Jtecx:
Quick update, this "fix" apparently also means NPC that you have yet to encounter also get added to the user's list, so....

Innoxia:
Thank you for this!

PR #1600, opened 2021-08-29
Content check

Maxis010:
Removed the need to add content checks to multiple locations when adding new content controlled fetishes

Added isContentEnabled to Fetishes and removed related checked in favour of isContentEnabled

No

0.4.1.5

Maxis010#3060


src/com/lilithsthrone/game/character/GameCharacter.java
```diff
@@ -6518,6 +6506,9 @@ public String addFetish(Fetish fetish) {
 	}
 	
 	public String addFetish(Fetish fetish, boolean shortDescription) {
+		if(fetish.isContentEnabled()) {
```
CognitiveMist: Should this be `!fetish.isContentEnabled()`?

src/com/lilithsthrone/game/character/fetishes/Fetish.java
```diff
@@ -30,10 +30,6 @@
  */
 public enum Fetish {
 	
-	// FETISHES:
-
-	// Sex types:
-
```
CognitiveMist: Is this removed intentionally?

src/com/lilithsthrone/game/dialogue/utils/PhoneDialogue.java
```diff
@@ -3632,10 +3621,14 @@ public DialogueNodeType getDialogueNodeType() {
 	
 	
 	private static String getFetishEntry(Fetish othersFetish, Fetish selfFetish) {
-		return "<div class='container-full-width' style='background:transparent; margin:2px 0; width:100%;'>"
-					+ getIndividualFetishEntry(othersFetish)
-					+ (selfFetish==null?"":getIndividualFetishEntry(selfFetish))
-				+ "</div>";
+		if(othersFetish.isContentEnabled()) {
```
CognitiveMist: So far, the others and self entries have always occurred in pairs, when it comes to fetish settings.

But what if in the future, we add content settings for the bimbo and crossdresser fetish, for example? (Or add new unpaired fetishes with settings.)

Maybe add a gate to `getIndividualFetishEntry` for when single fetishes are disabled, and also use a condition here like:
```java
if(othersFetish.isContentEnabled() && (selfFetish == null || selfFetish.isContentEnabled())) {
```

CognitiveMist (commented):
I like where this is going

src/com/lilithsthrone/game/character/GameCharacter.java
Maxis010: Yes it should be

src/com/lilithsthrone/game/character/fetishes/Fetish.java
Maxis010: I don't remember removing this but it's 3am
EDIT: Restored on local, will be added to next commit

src/com/lilithsthrone/game/dialogue/utils/PhoneDialogue.java
Maxis010: There are actually 3 fetishes that don't have pairings already, see `PhoneDialogue.java:3562-3573` of which 2 can be content blocked
Rather than make something complicated to address this for now I've left the existing code in place

CognitiveMist: I see, yeah, sounds good.

I have a few ideas for how to streamline this area a bit, but it can wait.

Maxis010: I have slept, and I have an idea on how to resolve this
Going to add pairings to the fetish object as they are used in a few locations (Phone dialog, Attacks, assigning fetishes at random) and then change the fetish menu at `PhoneDialogue.java:3541` into a loop so pairs and individual fetishes alike are handled cleanly without having to add them manually

Innoxia:
Thank you for this! 🎈

PR #1601, opened 2021-08-30
Several improvements around relationships

uglybead:

Relationship improvements (both speed and number of relationships understood).

1. Add additional direct parent/child relationships to those handled by the Relationship Enum.
2. Cache relationships between characters once computed (rather than redoing the graph walks).
3. Define relationship groups for use with pet names.
4. Use the relationship information directly to determine which "relationship pet name" to pick.

no

yes, against dev head.

n/a


src/com/lilithsthrone/game/character/GameCharacter.java
```diff
@@ -4004,7 +4004,21 @@ public void setGenericName(String genericName) {
 	public Map<String, String> getPetNameMap() {
 		return petNameMap;
 	}
-	
+
+	private static boolean containsAny(Set<Relationship> haystack, Set<Relationship> needles) {
```
CognitiveMist: This seems equivalent to `!Collections.disjoint(haystack, needle)`

Docs: https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#disjoint-java.util.Collection-java.util.Collection-

CognitiveMist (commented):
I would recommend invalidating a character's cache if any of the `setMother` or `setFather` variants is called.

Unfortunately, such a thing should probably also invalidate the caches of any other characters in their old and new lineages, too.

src/com/lilithsthrone/game/character/GameCharacter.java
uglybead: The setMother and setFather variants have two forms: one that is used during save loading and one that is used within an active game. Making the ones used for save-loading private should make the cache safe (for that function) since it isn't filled in unless requested (and we don't calculate family stuff during loading).

The other variants are used during initialization (either of the whole game or of the specific character for NPCChildren). We could (and probably should) make those versions protected. If we did, that change would remove the need to invalidate the cache, but fixing up the code in Game.java would make this PR rather large, so I've gone ahead and just invalidated the cache on that call for now.

Ultimately, lineages shouldn't need to be invalidated if we can isolate setMother/setFather to only be called during initialization (which should be possible).

CognitiveMist: During the course of the Main Quest, it is possible for Lyssieth to be your mother, and Lilaya & Meraxis to be your sisters. However, this is currently performed only within dialog nodes, and is not currently using the relationship system.

My concern is that future updates might change this, especially as more Lilin are added. Though, it is very possible that those will also be limited to the dialog system as well.

The other question is modding, as people have expressed interest in being able to modify relationship trees. However, I don't think that's a significant issue here, since saving+loading will refresh all realtionships.

uglybead: It sounds like what we need is two variants of the setMother and setFather functions: one which is cache-safe and is only used on newly created character (i.e., during game setup/loading or when creating a child) and a second set that can be used to "rewire the tree" (and thus flush the cache) in response to story events.

Would something along those lines be reasonable? If so, any opinions on what to call the cache-clearing variants?

PR #1602, opened 2021-09-01
WIP: Denial mod

ddmaker1234:

The currently existing denial content is very restrictive. This PR adds a few ideas that enable more creative incorporation of the fetish. There are also a few items and statusEffects that take advantage fo the new tags as an example of how they might be used.

Tag:
* PREVENTS_ORGASM - this can be added to a statusEffect to indicate that the given statusEffect makes it impossible to get off.
* ALLOW_ORGASM - a tag that can be added to a statusEffect as a counter to override the other mechanisms for denying a character. This allows for more permanent application of items and potentially temporarily overriding denial for limited periods.

Items (currently pills):
* Dose of Denial - an item inducing 7 days of denial
* Master of Denial - a more permanent form of the previous
* Dose of Release - an item overriding denial for roughly 1 hour
* Day of Release - an equivalent to Dose of Release but for 1 day
* Master's Antidote - an exceptionally expensive item that acts as an antidote to any of the denial items

Status Effects:
* Seal of Denial - The actual effect of the various denial pills
* Seal of Release - The actual effect of the various limited release pills 

ModNPC:
* Added code to allow the creation of an NPC purely in xml without having to modify java code.
* Added code to allow creation of an NPC trader purely in xml (though a bit crude at this point).

Side Note:
* Discovered and fixed bug in the way Quick Sex handles characters presently prevented from orgasming. That's the 76e48501d3d32a216468560c9c884828ef6bc70c commit and should be self contained.

In theory no. I tweaked some existing assets in the game to provide graphics for these, but that work is very crude. 

I've been lightly testing against 0.4.1.5 - commit 9f1354827

At present I don't have one ... This may be a one off may not be ... depends on how it's received.


ddmaker1234:
Add a fix for a bug in Quick Sex in 76e48501d3d32a216468560c9c884828ef6bc70c. I'm playing with this code applied to see if I spot any bugs/misbehaviors.  That particular commit should be self contained if it needs to be pulled out and rebased on to dev.

Balancing prices for the denial/release items may need some work. I think the release items are a bit too common/cheap.

ddmaker1234:
Second thought .. it looks like the stores don't actually do anything with rarity at all :(  ... might have to file a pr on that ..

CognitiveMist:
> Second thought .. it looks like the stores don't actually do anything with rarity at all :( ... might have to file a pr on that ..

This is intentional, and the presence of 'rarity' in the mod xmls is mostly a holdover from old versions, and that might be further pulled back over the next few updates.

The intended way, moving forward, is for nearly all items' rarity values to be determined by their effects, rather than their name/appearance/etc. For clothing, as an example, this means more enchantments -> more rare. For unique stuff, like the hypno-watch, special handling will be used.

ddmaker1234:
Ah ... so that's a derived value then. Ol ... I guess the only way to artificially restrict availability would be to limit who can sell things...

ddmaker1234:
Updated from dev ... I should probably go back through here and split out some of the code changes vs the xml stuff ..

PR #1603, opened 2021-09-29
Assorted Bugfixes

Maxis010:

Assorted bug fixes

Check commit messages

No

Yes, 0.4.1.9

Maxis010#3060, Phlarx#1765


src/com/lilithsthrone/main/Main.java
```diff
@@ -871,6 +862,24 @@ public static void saveGame(String name, boolean allowOverwrite) {
...
+			return true;
+		}
+		for(String c : "<>:\"/\\|?*".split("")) {
+			if(name.contains(c)) {
+				Main.game.flashMessage(PresetColour.GENERIC_BAD, "Incompatible characters!");
+				return true;
+			}
+		}
+		return false;
+	}
```
CognitiveMist, line 877: It is generally considered safer to have a list of allowed characters, instead of disallowed characters.

Also, maybe we should attempt to transform the name, instead of rejecting it, for example: a character name `Ra:t'hou Fh*a` might become the save name `Ra-t-hou_Fh-a`.... i.e. spaces become `_` and everything else becomes `-` so that the final allowed characters are `[A-Za-z0-9_-]`.

This is possible because the name conversion doesn't need to be reversible, only repeatable.

CognitiveMist (commented):
Another small bug appeared: https://discord.com/channels/302617912949211136/446038411791433729/893010795296874576

I think the fix may be:
```diff
--- a/src/com/lilithsthrone/game/character/body/Vagina.java
+++ b/src/com/lilithsthrone/game/character/body/Vagina.java
@@ -122,7 +122,7 @@ public class Vagina implements BodyPartInterface {

                descriptorList.add(Capacity.getCapacityFromValue(orificeVagina.getStretchedCapacity()).getDescriptor().replaceAll(" ", "-"));

-               descriptorList.removeIf(d->d.isEmpty());
+               descriptorList.removeIf(d->d==null || d.isEmpty());
                if(descriptorList.isEmpty()) {
                        return "";
                }
```

Nipples already have had this same change.

src/com/lilithsthrone/main/Main.java
Maxis010: The characters I'm rejecting are the danger characters for Windows, Mac and Linux (mostly windows)
From the comments in code I think it used to be an alphanumeric whitelist but was converted to a blacklist, I've just expanded on that

Innoxia:
Thank you for this!

PR #1605, opened 2021-10-20
Bugfix: Pattern id fixes

CognitiveMist:
Fix several issues around pattern modding:
- Fix bug where the ID of a pattern was inconsistent (e.g. "camo" versus "irbynx_camo"). This was fixed with duct tape for the built-in patterns, but modded patterns remained broken.
- Fix a bug where the svg defined in the pattern xml was not respected (would replace the xml file's extension only, instead of the full file name).
- Use full pattern id in saves, instead of just the name. Falls back to using just the name to load old saves, for compatibility.
- Update modding information for patterns.

Bug investigation started here: https://discord.com/channels/302617912949211136/446038411791433729/900455907371122729

Tested on: 0.4.1.9
Discord: Phlarx#1765


Innoxia:
Thank you for this!

PR #1607, opened 2021-10-25
Bugfix: Rare bug when importing clothing

CognitiveMist:
(probably could happen with other items, too, but reproduced when loading a save with clothes from an uninstalled mod within Nyan's 'special' inventory)

- On very rare occasions, `AbstractClothing.loadFromXML(e, doc)` was returning null, due to a null response from `ClothingType.getClothingTypeFromId(loadedId, slotHint)`, due to `Util.getClosestStringMatch(id, choiceMap.keySet())` returning a value that was not a member of the key set. Specifically, the value returned was the input.
- The bug is most likely due to the fact that the `getClosestStringMatch` function used a global variable, and so if the function was ever called in parallel, the two instances could and would interfere.
- As this is a race condition, it is very timing-sensitive, and so is both rare and unlikely to detect in the wild.
- Fix: refactor `getClosestStringMatch` to no longer use a global var, and adjust the consumers of that global var to use a different design.

Original report and investigation:
https://discord.com/channels/302617912949211136/446038411791433729/902082219059859456

Discovered on 0.4.1.1 with Java 8
Reproduced on 0.4.1.9 with Java 8
Tested on 0.4.1.9 with Java 8 and Java 14

Discord: Phlarx#1765

Note: additional safeguards may optionally be added somewhere to prevent the issue at other points in the chain (not an exhaustive list):
- In `Nyan.java` and `Monica.java`, within `loadFromXML`, add a check to prevent adding null AbstractClothing to the clothing lists. May also warrant the same for other merchants (these two are the only ones that use `getAllClothingListsMap`, so I didn't check the others).
- In `GameCharacter.java`, within `addClothing`, add a check for when the provided clothing is null (specifically in my case to prevent passing null through to `droppedItemText`). May also warrant the same for `addItem` and `addWeapon`.

### DIFF
--- a/src/com/lilithsthrone/game/character/attributes/Attribute.java
+++ b/src/com/lilithsthrone/game/character/attributes/Attribute.java
@@ -619,11 +619,7 @@ public class Attribute {
 			return oldConversionMapping.get(attributeId);
 		}
 
-		attributeId = Util.getClosestStringMatch(attributeId, idToAttributeMap.keySet());
-
-		if(Util.getLastStringMatchDistance()>3) {
-			return null;
-		}
+		attributeId = Util.getClosestStringMatch(attributeId, idToAttributeMap.keySet(), 3);
 		
 		return idToAttributeMap.get(attributeId);
 	}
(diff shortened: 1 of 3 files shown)


Innoxia:
Thank you for this!

PR #1610, opened 2021-11-01
Bugfix: sorting comparator inconsistency for horns and wings

CognitiveMist:
In the self-transform menu, the horn and wing types are sorted before displaying to the player.

The comparator used for sorting, in certain cases, could reveal itself to violate the general contract.

Since the sort function we are using is stable, the comparator given to it must also be stable. This is the general contract.

Stability primarily means that the sort order must be consistent (for example, if it says [1,2] is in the correct order, it must then say that [2,1] is in the incorrect order), and that it must not reorder equal elements (for example, it must not say that [1,1] is in the incorrect order).

The existing comparator function would fail this contract if HornType.NONE (or WingType.NONE) were ever passed in as the second element.

The rewrite changes from the two-argument comparator into the single-argument key extractor form, by using the Comparator class, and first-item enforcement (such as for HornType.NONE) is performed by mapping the desired first group to zero, and all other elements to one.

Bug report: https://discord.com/channels/302617912949211136/302618549820850176/904738543165923378

Tested on: 0.4.2
Discord: Phlarx#1765

### DIFF
--- a/src/com/lilithsthrone/game/dialogue/utils/CharacterModificationUtils.java
+++ b/src/com/lilithsthrone/game/dialogue/utils/CharacterModificationUtils.java
@@ -5,6 +5,7 @@ import java.time.format.TextStyle;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


Innoxia:
Thank you for this!

PR #1614, opened 2021-11-04
Enhancement: Add warning if weighted map contains bad values

CognitiveMist:
- Warn if all elements have zero weight.
  This will generally cause an unexpected null value, leading to an NPE.
- Warn if any elements have negative weight.
  This may or may not cause an unexpected null value, and may or may not cause other values to become impossible.

This effect isn't perfect, because it doesn't have a way to know which XML file a map might have come from, or similar feedback.

If there are specific problematic locations, `checkWeightedMap` can be used elsewhere with the `printWarning` parameter set to false, allowing the return value to inform other action.

If the global debug flag is set, then a stack trace is printed as well. No exceptions are thrown.

Tested on: 0.4.2
Discord: Phlarx#1765


Innoxia:
Thank you for this!

PR #1615, opened 2021-11-07
Enhancement: NPC Generation changes

AceXP:

Allow all NPC classes to be instantiated with the addNPC(String npcGenerationId, String parserTarget) method;

Added a fall-through method to instantiate the class (using the (NPC) Class.forName().newInstance() method). 

Now you can use one of the predefined names (like FieldsBandit), or use a qualified name (like dominion.DominionAlleywayAttacker) for any other NPC class.

Also updated the default (no parameter) constructors to make use of the gender preferences, rather than default values.

No.

Yes, tested with version 0.4.2.1 and the AceXP encounters mod.

AceXP#0930


Innoxia:
Thank you for this!

PR #1616, opened 2021-11-08
Addition: Add DamageVariance to the example CombatMove xml

CognitiveMist:
- damage variance was already implemented, but was not documented in hyena_bone_crush.xml
- also update doc links in dagger.xml to point at dev, instead of master

Tested on: 0.4.2
Discord: Phlarx#1765


Innoxia:
Thank you for this!

PR #1617, opened 2021-11-08
Enhancement: Dynamic tooltip heights

CognitiveMist:
The PR only applies to tooltips generated by `TooltipResponseDescriptionEventListener`, and so only affects the tooltips for the response buttons.

Further rollout can occur once this has been proven to be an effective and stable strategy.

This change also:
- Changed response 'description' css `height` to `min-height`, allowing the description box to grow.
- Combat moves which inflict multiple effects now display each effect on a new line.

In testing, I have noticed a few times where the first two or three tooltips generated can occasionally have a larger than expected height. After the initial few, all subsequent tooltips were fully stable. I am unsure of the cause.

Tested on: 0.4.2
Discord: Phlarx#1765

#### Images:
This tooltip would have overflowed, so that the lower text in the description would have been unreadable:
loooong

This tooltip is unchanged in shape and size from the current behavior:
normal

### DIFF
--- a/src/com/lilithsthrone/controller/MainController.java
+++ b/src/com/lilithsthrone/controller/MainController.java
@@ -2381,16 +2381,44 @@ public class MainController implements Initializable {
 			webEngine.loadContent(content);
 		}
 	}
-	
-	public void setTooltipContent(String content) {
+
+	/**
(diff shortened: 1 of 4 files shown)


Innoxia:
Thank you for this!

PR #1619, opened 2021-11-12
Bugfix: Stopped quick sex performing blocked actions

Maxis010:

Fixes quick sex performing actions that are not possible (i.e. penetrating sealed orifices)

Added check to foreplay and main sex preference to see if the original preference is possible
I've also bundled 2 small fixes

No

0.4.2.2 - Only tested in reported scene (Amber, PC Submissive)

Maxis010#3060


Innoxia:
Thank you for this!

PR #1622, opened 2021-11-20
Offspring seed

AceXP:
New PR for OffspringSeed

When loading an old save, any existing NPCOffspring NPC's that have not been encountered yet are converted to the new offspringSeed class. This is a one-time only event. It would be best to make a new save straight away. But also see the advise below.

Also some minor tweaks for dialogue when offspring is not birthed by the player, but by an incubator.

No

Yes, tested with 0.4.2.3 (and earlier versions). Although the recommendation is to do more testing.

So we have a better idea of who you are, what is your Discord Handle?
AceXP #0930


Innoxia:
Thank you for this!

PR #1624, opened 2021-11-22
Family class to centralise Litters

Maxis010:

Centralise the Litter objects rather than maintaining copies for each character

Added a new class (called Family but open to better names) to hold all of the Litter objects

No

Tested on 0.4.2.4

Maxis010#3060


Maxis010:
PR marked ready for testing purposes
Will make the class non-static so I can remove the fixes I've put in to stop litters being retained between saves

Issue #1538, opened 2021-04-16
The game is too easy


RandomNecromancer:
First of all, I will apologize because this topic will be quite large. Take your time to read and reply when possible.

Second, I understand that the game is still in development and that these problems probably arose for that reason, but since this game is open for contributions I decided that it might be a good idea to share the experiences I had while playing it, despite risking stating the obvious.

The biggest issue in this game at the moment is that it is too easy even (sometimes especially) in higher difficulties. The two main causes are the ease in getting rich and reaching the maximum level. There are several ways to achieve these goals, but I will describe only the one I use, as I think it is one of the most efficient:

1. After starting the game, finish Lilaya's initial interactions (tests, current date, enchantments ...) and you should have enough points to get the "orgasmic level drain" perk.

2. Get the prostitution license and work at the brothel, "grinding" your way up to the max level using quick sex and earning some money in the process that will be needed later. Especially easy and fast in higher difficulties and with the right combination of shortcuts. It never takes more than a few minutes to finish.

3. Get the enslavement license at the slaver's alley and set up a slave room and a milking room (fully upgraded) and remember to activate the auto sell options and assign the slaves later.

4. Enslave a slime and, through the transform option, set it to have penis, vagina and the maximum amount of boobs and crotch boobs, as well as maximum cum and milk "storage" and regeneration. Skip some days through sleeping and collect your money.

It may seem like a lot of work, but considering that you haven't done any major campaign missions yet, you can just save and never do it again. Want to start a new character? reload that save and that's it, you are already maxed and you can play the whole campaign again in an almost completely different way.

I don't believe it is possible to solve these problems because the very concept behind orgasmic level drain and milking rooms is faulty, but it is possible to mitigate at least. Just think about it:

1. Why isn't everyone super strong if it's so easy to gain a level and get the orgasmic level drain perk?
In a world where sexuality is so open, I can't think of any plausible reason, even considering the fear that people feel about working in a brothel sanctioned by enforcers (of course, not everyone would even be willing to work in a brothel, but you know what I meant). In the player's case, it is not a big problem in the standard difficulty, since the npcs do not have their levels scaled with the player, but the same does not apply to higher difficulties.

2. Why aren't all slavers extremely rich (are they? I don't remember the game stating anything about it) when it is so easy to enslave a slime and sell its fluids?
I know it is difficult to get a license, but I am only referring to those who already have one. Also, I didn't mention demons for lore reasons.

[...]

PR #1626, opened 2021-11-23
Milk Tax

Maxis010:

Balance the milking room against the other jobs and fix a dupe bug

Reduced fluid values and increased upkeep

No

0.4.2.4

Maxis010#3060


CognitiveMist:
There is a restriction that the player may only build one spa; would it be worthwhile to restrict the number of milking parlors the player can build? To 4, for example.

(I don't know how many milk rooms are used for a max-capacity milk farm currently.)

I feel like selling all fluids for the same amount is rather bland (even though I usually deal with 'normal' attributes/flavors, myself). Perhaps leave the flavor adjustment in, but take the texture stuff out. Or give a flat boost for non-standard flavor, and another for non-standard texture. Of course, the boost can be much smaller than the existing 50%.

Another idea I had (this would probably fit better into an economy-based nerf, though -- meaning market saturation, etc. -- and also probably require a nontrivial amount of writing):
Change the texture/flavor boosts over time. On one week, honey flavor might have a 20% boost, and the next 0% (meaning it's irrelevant), and the next week it might actually have a 20% penalty. Allow the player to read the market (maybe with upkeep just for being able to do that). This way, a max-efficiency farm needs a lot of player intervention to run at peak profit. If left unmanaged, the profits would be roughly in line with having fully-standard fluids. Of course, the static adjustments in this PR would still be beneficial.

defsil:
> (I don't know how many milk rooms are used for a max-capacity milk farm currently.)

_One slave is all it takes_.

image

And this, of course, is not _nearly_ the max capacity, nor the maximum possible milking profit per slave. It's just that broken. Having to build and fully populate many milking rooms to turn such a profit, on the other hand, would be exactly the intended solution and a great way to earn flames, so there's no point in limiting _that_.

Also, please have a look at [this issue](https://github.com/Innoxia/liliths-throne-public/issues/1538) and the discussion of milking in the comments. :3

defsil:
I must say that, as much as I like the idea of rebalancing milking, perhaps lowering the affection values for it and completely removing flavor & texture influence on the cost is just a bit too much? 
Also, "Increased affection penalty for overworking" should definitely get split into a separate pull. :&)

Maxis010:
> I must say that, as much as I like the idea of rebalancing milking, perhaps lowering the affection values for it and completely removing flavor & texture influence on the cost is just a bit too much?

I'm not lowering the affection of milking, I'm lowering the bonus from the fetishes as with 24 hour milking you can go from -100 to 100 in 4 days, the standard affection rate has not been changed

Modifiers as written was broken, you just add all the modifiers to min/max the price. Without factoring in modifiers regardless of what race goes in the outcome is the same. Flavour could be restored with minimal impact, I removed all modifiers for testing and personally see no reason it needs to be readded.

> Also, "Increased affection penalty for overworking" should definitely get split into a separate pull. :&)
Affection penalties are a part of this rebalancing as there is no real penalty for overworking, even if the min/maxers won't care.
If anything from this PR should be split it's the bug fix I've bundled inside
That said Inno will go over this code with a fine tooth comb and if she has an issue I can easily split the 3 parts of this PR for her.

> There is a restriction that the player may only build one spa; would it be worthwhile to restrict the number of milking parlors the player can build? To 4, for example.
> 
> (I don't know how many milk rooms are used for a max-capacity milk farm currently.)

As defsil has said, 1/8th of a room is all it takes
 
> I feel like selling all fluids for the same amount is rather bland (even though I usually deal with 'normal' attributes/flavors, myself). Perhaps leave the flavor adjustment in, but take the texture stuff out. Or give a flat boost for non-standard flavor, and another for non-standard texture. Of course, the boost can be much smaller than the existing 50%.

I removed all of the modifiers for testing purposes, there is no reason that the 50% flavour boost shouldn't be put back, but I will leave that to Inno to decide

> Another idea I had (this would probably fit better into an economy-based nerf, though -- meaning market saturation, etc. -- and also probably require a nontrivial amount of writing): Change the texture/flavor boosts over time. On one week, honey flavor might have a 20% boost, and the next 0% (meaning it's irrelevant), and the next week it might actually have a 20% penalty. Allow the player to read the market (maybe with upkeep just for being able to do that). This way, a max-efficiency farm needs a lot of player intervention to run at peak profit. If left unmanaged, the profits would be roughly in line with having fully-standard fluids. Of course, the static adjustments in this PR would still be beneficial.

Great idea, but my goal was to counter the 1 million flame/day slaves not build a market sudo-simulation

CognitiveMist:
> One slave is all it takes.

> As defsil has said, 1/8th of a room is all it takes

I fail to see how this is relevant to the suggestion of limiting the number of milking rooms. Of course it won't fix the problem on it's own, nobody suggested that.

Two slaves is always gonna be twice as profitable as one; 64 slaves twice as profitable as 32. If someone had a single full room, and their profits got cut to 10%, they can just build and fill 9 more rooms to get the same profit. A bit more effort? Yes. Same absurd profits? Also yes.

I know I'm drastically oversimplifying, but I'm trying to point out that limiting the number of rooms is also an effective way of limiting profit. Especially since buying/building additional property is still on the cards for future updates.

Innoxia:
Thank you for this!

PR #1630, opened 2021-12-24
Bugfixes & QoL changes

Maxis010:
Assorted bugfixes

> Changed NONE type penis & vagina to non 0 default values

Default values should respect the race default values

> Cap the event log at 50 entries

`EventLog` was uncapped (memory leak?) but only 50 entries were saved to file

> Stop gear sealing onLoad if it was unlocked
> Dialog / Tooltip fixes

Self explanatory

> Moved QUEST rarity check to banishNPC

Original check in `unequipClothingIntoVoid` causing duplication bug

> Changed all removeNPC to banishNPC so that all checks are applied

Ensure all NPCs are subjected to the banishNPC checks before being removed

No

Ongoing testing on 0.4.2.5

Phlarx#1765 - Improved Parser error message
Maxis010#3060 - Everything else


Maxis010:
> Fixed bugs with event log from capping at 50 entries

During testing I found the event log was not consistent between saving and loading
Part of this is related to the sorting performed on load where the constructor was setting in seconds but most events were being set in minutes
Also multiple events occurring at the same time (such as when trading with a shop) were not ordered correctly
Removed the time from the log as it was only being used as part of this sort, and I found a better solution as the order is now preserved

Innoxia:
Thank you for this!

PR #1632, opened 2022-01-02
Combat rework & re-balance

nerdlingerLT:

To re-think the way combat is approached as to prevent mindless spamming of powerful moves and increase the viability previously less optimal playstyles.
Not yet implemented: allow weapon damages to scale and retain their usefulness as the character progresses, reducing the amount of new weapons needing to be added every time the story reaches a new area.

Many changes to many Boss encounters
Combat moves and Spell damage now scale with Character stats.
Many weapons have been given new damage values and effects.
New stat values for a few Jobs and Perks.

So far, no.

Last tested on 4.1.1.

nerdlinger


CognitiveMist:
This change includes a bunch of svg and xml edits, but I don't see any java changes, so some things seem to be missing? (Boss changes, damage scaling, job/perk changes, etc)

Looking at the branch history, it seems to be based on master, when it really should be based on dev (so that it is incorporated with the latest game code).

BTW, if this is a work in progress, you could set it to be a draft under the Reviewers section in the right column. This signifies that it's not ready to go just yet, and that you are still working on it.

PR #1634, opened 2022-01-15
BugFix: Determine species name with Body instead of GameCharacter

AceXP:

In the phone menu and elsewhere, species names (like horse-morph, cat-morph etc) need a GameCharacter to show all variations (taurs, demonics, lamias etc). And because an OffspringSeed isn't a GameCharacter this does not work and only simple names are shown. This PR fixes this by substituting the Body for the GameCharacter.

Refactor the relevant methods to use a Body parameter instead of the GameCharacter parameter. Use GameCharacter.getBody() where needed to supply the parameter.
Also, slimes need the fleshSubspecies attribute, this was therefore moved from GameCharacter to Body. Again use GameCharacter.getBody().getFleshSubspecies where needed.

No.

Yes, tested with version 0.4.2.9

AceXP#0930


Innoxia:
Thank you for this!

PR #1638, opened 2022-01-18
Bugfix: Allow orgasm actions while player is immobilized

miome:
- As of 4.2.9, when a player is immobilized then no orgasm options are displayed.  The player is left with a blank grid with no options, and has to load from a previous save.
- This modification allows Prepare and Orgasm options to be displayed
- Deny option is deliberately excluded
- Changed tested and working on my system with latest dev 4.3.0 branch and with the release 4.2.9 branch.
- My discord handle is Miome#2106


Innoxia:
Thank you for this!

PR #1639, opened 2022-01-20
Buggy menu custom sex scene creator

Amarok909:

Adds new feature to buggy menu to allow sex with any character present. Added because some NPC sex scenes have limitations (post-Arthur Zaranix and Vicky are sub only) and allows larger orgies that can include unique NPCs.

Added a page in the buggy menu that allows the creation of custom sex scenes. Participants could be picked from those in the same room as the player. Consent and Sub-control can also be controlled.

No

Yes. v0.4.3

Amarok#6129


Amarok909:
Older saves which don't have XML sections for each of the Ellis characters and Meraxis, can have issues.
This can be solved either by starting a new game OR somehow adding sections for the unique characters to the old save-game.
Is there a method of automatically adding missing characters to the save file?
EDIT: seems there are issues with the initUniqueNPCs() method. the root issue seems to be in GameCharacter

Gamecharacter: ~line 556
`this.setLocation(
				worldLocation, 
				Main.game.getWorlds().get(worldLocation).getCell(startingPlace).getLocation(),
				true);
`
getCell() is returning null, verified with following code
`getCells(startingPlace).size()` which returns zero in Takahashi's and other character's cases

EDIT2: GameCharacter setLocation issues were caused by WorldTypes not being loaded properly. I've fixed this issue in #1640

PR #1640, opened 2022-01-21
World Generation auto-repair functionality

Amarok909:

Adds a diagnostic/automatic repair feature to World Generation when loading an old or slightly damaged save file

When loading a save file, the version checks on new vanilla worlds are only done once rather than looped. After the null world check, a new piece of code has been added to verify cell-to-world associations; if getCell(PlaceType) comes back as null, then that world will be regenerated.

No

Yes. v0.4.3

Amarok#6129

### DIFF
--- a/src/com/lilithsthrone/game/Game.java
+++ b/src/com/lilithsthrone/game/Game.java
@@ -936,72 +936,85 @@ public class Game implements XMLSaving {
 				}
 				
 				// Add missing world types:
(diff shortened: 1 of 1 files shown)


Amarok909:
marked WIP to fix a found bug and add better functionality

CognitiveMist:
Github allows setting a PR to be a draft (which is inherently WIP).

This option is found in the right column under the 'reviewers' category. A bit hidden, but it is what it is.

PR #1643, opened 2022-01-24
small suggestion

haoknowah:
right click function
Can it be made so that the right click does something, like maybe a quick shortcut if used on items in inventory so you don't have to select it and then select the option you want? Maybe use it to quickly deal with clothes in sex scenes or consume one of an item in inventory?


haoknowah:
Comments would be welcome.

PR #1665, opened 2022-04-23
Slime Tweaks

MintyChip:
Tweak how slimes work a bit.

1. made it so that the slime attacker's slime transformation applies the same effects as the biojuice for the sake of consistency
2. Added flavor text to the biojuice description to mention that it's "heavily processed" and thus corruptive where the source (slime queen in the bath scene) is not
3. Nerfed the permanent corruption gain from biojuice, as it seemed like a harsh effect if a new player just wandered into the submission.
4. Made it so that elementals can't have their body material transformed into SLIME or FLESH using the biojuice or flesh potions.

No
0.4.3.9
MintyChip#1944


Innoxia:
Thank you for this!

PR #1675, opened 2022-06-22
Crossdressers get a different outfit from Brax

Nonny5Mouse:
- What is the purpose of the pull request? (To let players wishing to use opposite femininity clothing to get a differing enforcer set)

- Give a brief description of what you changed or added. (The check to decide which set is obtained now checks if the player has the crossdresser fetish. Masculine crossdressers get the female set and female crossdressers get the male set.)

- Are any new graphical assets required? (No)

- Has this change been tested? If so, mention the version number that the test was based on. (Yes, 4.4.1. Tested with masculine and feminine characters with and without the fetish)

- So we have a better idea of who you are, what is your Discord Handle? Nonnymouse#5520


CognitiveMist:
The condition could be simplified by using xor (`^`), but clarity of intent is the most important thing.

Nonny5Mouse:
Extras added:
-Fixed drake-description referring to wyvern-morphs
-Made the ryu subspecies give an exception to the tufted tail if the character has a serpentine lower body.

CognitiveMist:
I really like the core idea here (and it's a really simple addition), but I just had a thought:
Why not just have two buttons? One to take a masculine uniform and one to take a feminine one. (With a lore text to say it's too risky to grab both, or something).

PR #1683, opened 2022-08-04
Busting duplicate line <3

IjyiH:
*blatantly disregards PR template*

That line was assigning a settings value, twice. Harmless to leave there but the code looks prettier without doing extra work.

I got rid of Discord, haven't logged in for years. Glad to see the project's still alive :D

cheers, IJH


Innoxia:
Thank you for this! Good to see that you're still around!

PR #1588, opened 2021-08-06
Include JavaFX and Nashorn in the JAR so we're not stuck with Java 1.8 forever

ElAcechador:
- What is the purpose of the pull request?

Fixes #1435 and #1546 (among others) and removes the need for this part of the readme:

https://github.com/Innoxia/liliths-throne-public/blob/c863c6d5ad1ed625200b16791c4c707d57aa4e03/README.md#L7

- Give a brief description of what you changed or added.

I added the required JavaFX dependencies to pom.xml and [followed the steps](https://openjfx.io/openjfx-docs/modular#Non-Mod) needed to bundle it inside the JAR using the maven-shade plugin. JavaFX does have a bunch of native components that are OS-specific and I'm bundling Windows, Mac and Linux, so that currently bumps the JAR size from 20 MB to 120 MB. I've also replaced the Oracle JDK Nashorn with the OpenJDK version now that it's been split off into a separate library.

The few other changes I made, like adding the Launcher class were needed to get the game to run as a standalone JAR. [JavaFX doesn't like it when your main class doesn't inherit `Application`](https://bugs.openjdk.java.net/browse/JDK-8093075?focusedCommentId=13756386&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13756386), but the dependencies bundled in your JAR can't be required by the main class directly.

(for the record, I haven't done Java in a good 5 years and I have never seriously used  Maven before)

The build tutorial might need to be updated in a few places. The game can be built with a simple `mvn compile` and `mvn package` (although `mvn clean` might be needed at first if this gives you compilation errors), but I haven't tried the other suggested ways (though I would assume that those would still work for the most part, given the old main function is still there).

- Are any new graphical assets required?

No

- Has this change been tested? If so, mention the version number that the test was based on.

Not beyond checking that it runs and skimming through the prologue with the changes.

- So we have a better idea of who you are, what is your Discord Handle?

I'm not currently on the Discord server.

PR #1684, opened 2022-08-09
Feature/java 17 + visual fixes + release pipeline

Tazze:

Make LT runnable directly on the most recent Java LTS (17) and create an automatic build and release pipeline to reduce workload.

  - Added dependencies for Nashorn and JafaFX which are no longer part of the JDK
  - Added a launcher since non modular (non-jlink) builds can no longer run JavaFX Applications directly
  - Made it so that the WINDOW_IMAGE is not set before JavaFX initializes itself, which causes an exception
  - Use Maven Shade plugin to build the runnable jar
  - Fixed elements that are supposed to be transparent being opaque (for example: disabled buttons)
  - Add a pipeline to compile the game for all the different platforms and architectures as a sanity check
  - Add a pipeline to create a release automatically every time a new tag is pushed

No.

Not fully tested on version 0.4.4.5 (game runs, able to start a new game, create a new character and skip the prologue without issues)

Tazze#9319

NOTES:
I have decided to add "(WIP)" until I'm sure there are no regressions associated with these changes
How to build and run (linux):
`mvn clean package`
`cd target/Lilith's Throne (linux)`
`java -jar "Lilith's Throne-0.4.6.jar" `

How to build and run (windows)
`mvn clean package -Dtarget.platform=win -Dexclude.platform=linux`
`cd .\target\Lilith's Throne (win)`
`java -jar "Lilith's Throne-0.4.6.jar"` (alternatively, double click the jar file)


ElAcechador:
See #1588, which is very similar but builds separate builds per platform, among other things.

IIRC this has languished because Innoxia doesn't actually use Maven (I might be wrong, someone please confirm), so it wouldn't be useful to generate the actual builds, but you seem to be using github CI, so that might actually be different.

Tazze:
> See #1588, which is very similar but builds separate builds per platform, among other things.
> 
> IIRC this has languished because Innoxia doesn't actually use Maven (I might be wrong, someone please confirm), so it wouldn't be useful to generate the actual builds, but you seem to be using github CI, so that might actually be different.

My release pipeline is designed to build jars per platform and architecture (for Apple Silicon Macs, x86 Windows and ARM64 Linux, on top of AMD64 builds for all three), see https://github.com/Tazze/liliths-throne-public/releases/tag/v0.4.4.5-hotfix-2 (since you seem to imply my PR doesn't)
I also don't see why maven wouldn't be useful, the jars generated work just fine and are what I use to play the game, since I refuse to install an ancient JDK like 8

This PR also fixes a bunch of visual bugs that occur on current versions of JavaFX because of a slight change in CSS rendering behavior

zqqui:
this update needs to be merged. jdk8 is decrepit

Innoxia:
Thank you for this!

PR #1688, opened 2022-08-17
Specific lilin parentage flags

MintyChip:

Add specific flags to indicate the pc's lilin heritage rather than just checking if they are a demon

Added and implemented a "child of lyssieth" flag to innoxia/flags.xml 
Added getLilinMother() method to the player character class.

No
0.4.5
mintychip#1944


Innoxia:
Thank you for this! 😈

PR #1695, opened 2022-08-30
Revaluation of enchantments

Maxis010:

Balance the value modifiers for items, clothing and weapons

Changed `getValue()` of AbstractItem, AbstractClothing & AbstractWeapon
Removed any unique valuations

No

0.4.5

Maxis010#3060


Phlarx:
If I'm reading this correctly, an shirt with 50 "+1 physique" enchants will be worth *less* than a shirt with a single "+3 physique" enchant. Is that accurate?

Maxis010:
> If I'm reading this correctly, an shirt with 50 "+1 physique" enchants will be worth _less_ than a shirt with a single "+3 physique" enchant. Is that accurate?

Yes, rather than increasing the value for every enchantment it just takes the strongest enchantment as a multiplier

CognitiveMist:
> > If I'm reading this correctly, an shirt with 50 "+1 physique" enchants will be worth _less_ than a shirt with a single "+3 physique" enchant. Is that accurate?
> 
> Yes, rather than increasing the value for every enchantment it just takes the strongest enchantment as a multiplier

Hm, ok. Doesn't make much sense to me, but as long as it's known.

Maxis010:
Considering changing this to factor in number of enchants but at a lesser rate

Maxis010:
> If I'm reading this correctly, an shirt with 50 "+1 physique" enchants will be worth _less_ than a shirt with a single "+3 physique" enchant. Is that accurate?

This is no longer the case, buffed all of the existing modifiers and restored a number of enchants bonus

Innoxia:
Thank you for this! ⚔️✨⚖️✨👗

PR #1706, opened 2022-09-19
Bugfix: Normalize the default crit logic

CognitiveMist:
The default crit logic (crit on third use) was a bit ambiguous and unpredictable. This change fixes that.

For most players this logic change will go unnoticed, but I did adjust the default crit text to clarify that it is per-turn.

Below I have an example of what the issue is. I gave myself 10 AP for ease of communication. With 3AP, the difference is nothing, but if a player ever gets an effect that adds even 1 additional AP, then the issues could appear.

This change only affects moves that use the default crit logic; any moves that define their own logic are unaffected.

No new assets required.
Tested with version 0.4.5.5.

## Old behaviour:
Queued (2 attack crits, 0 block crits):
image
Executed (3 attack crits, 0 block crits):
image

## New behaviour:
Queued (1 attack crits, 2 block crits):
image
Executed (1 attack crits, 2 block crits):
image

### DIFF
--- a/src/com/lilithsthrone/game/combat/moves/AbstractCombatMove.java
+++ b/src/com/lilithsthrone/game/combat/moves/AbstractCombatMove.java
@@ -9,6 +9,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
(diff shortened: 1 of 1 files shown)


Innoxia:
Thank you for this!

PR #1711, opened 2022-09-30
[Bugfix] Command_unknown in sex text

Nemiander:

Command_unknown error in starting dialogue from the spitroasted(front) with one occupant offspring and one non-offspring. This option is available with companions turned on.

Seemed that "your" was accidentally tried as a command and so changed it to just text.
no
0.4.6
Duskit#6985
First time i did something like this so I'm a bit nervous on whether i did this pull request right.


Innoxia:
Thank you for this!

PR #1712, opened 2022-09-30
Add regular arms and back wings to raptors

DSGMods:
Adds regular arms and back wings to the raptor race

Added two new bodyparts to the raptor race

None

Tested on 0.4.6

dsg #4913


Innoxia:
Thank you for this!

PR #1713, opened 2022-10-02
Controller Rework & UI Changes

Maxis010:
Reorganise tooltip and click listeners
Standardise what is and isn't shown in transformations menus based on content options
Clean up some minor UI bugs

Removed `MainControllerInitMethod` by spreading the contents across other more specific controllers
Removed content option paging, instead making them separate dialogue nodes
Added disable states for multiple transformation options

No

All tested against 0.4.6

Maxis010#3060


src/com/lilithsthrone/game/character/GameCharacter.java
```diff
@@ -26915,12 +26915,21 @@ public String incrementBreastStoredMilk(float increment) {
...
 	public FluidRegeneration getBreastLactationRegeneration() {
+		if (!Main.game.isLactationContentEnabled()) {
+			return FluidRegeneration.ZERO_SLOW;
+		}
 		return body.getBreast().getLactationRegeneration();
 	}
 	public int getBreastRawLactationRegenerationValue() {
+		if (!Main.game.isLactationContentEnabled()) {
+			return 0;
+		}
```
CognitiveMist: Unsure if the following is acceptable or bad (though the situation may be rather unlikely):
1. Start game or load game with lactation content enabled
2. Set lactation values to an amount
3. Disable lactation content
4. Trigger a call to `incrementBreastLactationRegeneration`
5. Reenable lactation content
6. Lactation values will be strange because (4) incremented form zero, instead of from current value

True for cum storage, and crotch milk regen, and anything else whose `increment*` uses the getter.

src/com/lilithsthrone/controller/CoveringController.java
```diff
@@ -0,0 +1,461 @@
...
+import com.lilithsthrone.main.Main;
+import com.lilithsthrone.utils.colours.Colour;
+import com.lilithsthrone.utils.colours.PresetColour;
+
+public class CoveringController {
+	private static final boolean noCost = !Main.game.isInNewWorld()
+			|| ScarlettsShop.isSlaveCustomisationMenu()
+			|| Main.game.getCurrentDialogueNode() == MiscDialogue.getMakeupDialogueForEqualityCheck()
+			|| Main.game.getCurrentDialogueNode() == RoomPlayer.AUNT_HOME_PLAYERS_ROOM_MAKEUP
+			|| Main.game.getCurrentDialogueNode().getDialogueNodeType() == DialogueNodeType.PHONE;
```
CognitiveMist: This looks wrong; was this variable supposed to have been inside (or a function called from) each of the `init*Listeners`?

src/com/lilithsthrone/game/character/GameCharacter.java
Maxis010: Will review if it is possible to call `incrementBreastLactationRegeneration` while lactation is disabled. I wouldn't expect it to be but then this is written in spaget

src/com/lilithsthrone/controller/CoveringController.java
Maxis010: `noCost` has not been changed as part of this rework
[See here](https://github.com/Innoxia/liliths-throne-public/blob/dev/src/com/lilithsthrone/controller/MainControllerInitMethod.java#L4936)
It is checked about 2 dozen times in various listeners, could be made into a method call

src/com/lilithsthrone/game/character/GameCharacter.java
CognitiveMist: 'Possible' includes mods, unfortunately, so yes it is possible.

src/com/lilithsthrone/controller/CoveringController.java
CognitiveMist: I would consider this to be changed, because it has been moved from inside a function, to outside any function; instead of the value being set once per dialog, it is set once.

Maxis010: Changed to method call

src/com/lilithsthrone/game/character/GameCharacter.java
Maxis010: These checks were made redundant by changes to the UI so I have removed them

Innoxia:
Thank you for this!

PR #1716, opened 2022-10-09
Import modded character

AceXP:
 
Add the possibility to load character .xml files when modding the game.

A NPC .xml file can be loaded back into the game as a 'moddedCharacter', which can then be used in an encounter. There is an example included, an encounter with a shady dealer in the back alleys of Dominion.

An explanation of how this works was added to modding.txt

No

Yes, tested with 0.4.6.3 (and a bit)

AceXP#0930


Innoxia:
Thank you for this! 🦄

PR #1721, opened 2022-10-13
added the ability to choose which outfit you take from Brax's office.

Nonny5Mouse:
To make it so players can obtain the outfit of their choice, regardless of if they're feminine or not.
(Edit: Also, if the debug menu is used to obtain a uniform, it'll give both sets, but only one belt and jacket.)

Changed the BraxOffice.java file and more specifically the function that gives players a uniform, had to make slight alterations to brax.xml and Quest.java

I tested it on 4.6.5, though I typed the wrong version number on the two java files I altered. I tested the bluff, wolf tease(sub and dom), and fight options.

Nonnymouse#5520


Innoxia:
Thank you for this! 🐺

PR #1724, opened 2022-10-18
Show when a NPC is winged, if its race is usually not winged

Stadler76:
Added 'winged' to NPC lists and tooltips for races that usually have no wings.

Thankfully adding generic wings to an NPC or the player, that is of a race that usually has no wings it's not considered 'lesser' anymore, but a winged wolf-morph for example still shows up as a wolf-morph. I've changed that.

Screenshots:
Tooltip and characters present Tooltip and occupant list

No

Yes, 0.4.6.6 Alpha

Stadler#3007


Innoxia:
Thank you for this! 🐦

PR #1742, opened 2022-12-01
Enable Tattoo Rendering with Non-Standard Colours

NoHornyOnMain:
To allow rendering of tattoos using non-standard colours (e.g. via save editing).

Removed two checks that cause tattoo rendering to silently fail if the primary colour is not in that tattoo's usual set of primary colours, even when it would have otherwise succeeded.

Actual rendering errors will still print the stack trace to the log and return an empty string.

No, this doesn't require any new graphical assets.

Yes, it's been tested on version 0.4.7 (specifically commit 4e645da98afdd6189f4eb52d46f04e3391e4f58f).

NoHornyOnMain#5417


Innoxia:
Thank you for this! 💯

PR #1745, opened 2022-12-04
Improve Maven Compatibility Across JDK Versions

NoHornyOnMain:
It's currently impossible to compile from source without changing files (pom.xml or UtilText.java) unless you use a non-Maven compilation method (i.e. Eclipse) on JDK8. As such, this PR aims to allow Maven-based compilation on any reasonable Java version without requiring any changes.

- JDK version is now checked when compiling with Maven (even via a supporting IDE).
- JDK8 specific pom.xml changes are applied if needed (e.g. maven.compiler.source/target).
- JDK8-style internal Nashorn imports are swapped out at compile-time if needed, then swapped back afterwards.
- External JavaFX and Nashorn dependencies are only downloaded if running JDK11+.
- Provides explicit support for all LTS JDK versions (8,11,17) and implicit support for JDK11+.

No, this doesn't require any new graphical assets.

This has been tested and confirmed working with the following setups:
- Maven
    - JDK8
    - JDK11
    - JDK17
- IntelliJ IDEA
    - JDK8
    - JDK11
    - JDK17
- Eclipse Oxygen
    - JDK8

All of the above testing was performed using version 0.4.7 of the game (specifically commit 4e645da98afdd6189f4eb52d46f04e3391e4f58f).

NoHornyOnMain#5417

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,53 +69,65 @@
 				<maven.compiler.release>17</maven.compiler.release>
 			</properties>
 		</profile>
+		<profile>
+			<id>external-javafx</id>
+			<activation>
+				<jdk>[11,)</jdk>
+			</activation>
+			<dependencies>
+				<dependency>
+					<groupId>org.openjfx</groupId>
+					<artifactId>javafx-base</artifactId>
+					<version>${javafx.version}</version>
+					<classifier>${target.platform}</classifier>
+				</dependency>
+				<dependency>
+					<groupId>org.openjfx</groupId>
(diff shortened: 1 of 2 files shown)


CardcaptorRLH85:
I've just inadvertently tested this on JDK 19.0.1 since I didn't notice that Chocolatey had updated me away from the v17 LTS.  It built just fine for me before and after making my own #1749 PR edits.

Innoxia:
Thank you for this PR! Sorry that it took me so long to merge!

PR #1749, opened 2022-12-24
Bump pom.xml version tag

CardcaptorRLH85:
I've updated the pom.xml version number to 0.4.7, the number of the current public release.

Should the version tag in pom.xml be bumped for each version?  If so, this PR does it for this release but it should be added to the release process for each version moving forward.  I only noticed when I built it from source myself and it output Lilith's Throne-0.4.6.3.jar.


Innoxia:
Thank you for this, but as LT is now on v0.4.8.1, I've updated the version to that instead. I'll try to remember to keep this version number updated in the future!

PR #1751, opened 2023-01-01
Seal removal tweak

RubberSiev:
Simple tweak to make the seal removal system somewhat less exploitable by making it so creating clothing with servitude + seal on it doesn't give the player a free pass to get out of all bound items.

Pretty self explanatory. Change has not been tested because I don't have a Java environment set up to do it. Discord tag is Siev#4561 if you wanna yell at me for doing it wrong, my Java is very rusty =p


Innoxia:
Thank you for this PR, however I think I'll leave the unsealing behaviour as it is, due to the possibility of additional sealed clothing preventing removal of the clothing which Lilaya unseals, which would leave the player in a stuck state. e.g. Lilaya unseals underwear that prevents the player from unsealing clothing, but sealed shorts would prevent the player from taking them off, so they'd be stuck. 💦

PR #1752, opened 2023-01-16
Bugfix: intermittent failure on "New Game"

CognitiveMist:
A random failure could occur due to the place count map being static, but not thread-safe 

- Switched to concurrent map for place counting
- Moved place counting after null check (necessary for ConcurrentHashMap usage)
- Added try/catch around world type construction (to reveal any other potential errors in the area, while preventing a soft-lock)

Tested on a dev build of the 0.4.7 release
Discord: Phlarx#1765


Innoxia:
Thank you so much for this PR! It would have taken me a very long time to figure this one out, so this was a really big help!

PR #1754, opened 2023-02-26
Calculate subspecies weights within race first

SightglassCat:
This pull request modifies "getSubspeciesFromBody" in AbstractSubspecies to calculate subspecies weights only within the NPC's race. This makes loading save files significantly faster, especially for large saves in games with many modded subspecies, since fewer subspecies weights need to be parsed and calculated for each NPC. 

If the modified loop fails to find an highest weight among the race's subspecies, it will calculate weights for all subspecies, as in the original code. 
No new graphical assets are required.

This code has been tested for the following versions:
- 0.4.7
- 0.4.7.5

The speed improvement was tested using a save file with ~9500 npcs and 204 subspecies. The time required to load the file using the unmodified game was 7 minutes, while the modified game took about 40 seconds.

Discord ID: Sightglass#1408

### DIFF
--- a/src/com/lilithsthrone/game/character/race/AbstractSubspecies.java
+++ b/src/com/lilithsthrone/game/character/race/AbstractSubspecies.java
@@ -828,7 +828,7 @@ public abstract class AbstractSubspecies {
 		
 		int highestWeighting = 0;
 		int newWeighting;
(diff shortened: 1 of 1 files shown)


Innoxia:
Thank you very much for this! 💯

PR #1759, opened 2023-03-16
Add micro bikini pattern support

DSGMods:
Adds pattern support to both micro bikini items

Added patternLayer group to both items
Changed two paths to a compound path in the bikini top

Included and optimized

Tested in 0.4.7.5

dsg#4931


Innoxia:
Thank you for this! 👙

PR #1762, opened 2023-03-21
Fix to missing ID and path for anonymous-classed NPCs

SightglassCat:
Adds checks for anonymous classes when creating and saving a npc.
Added check for anonymous class in:
GameCharacter.saveAsXml for the pathName node
Game.addNPC when setting an NPC's id
None
Tested on v0.4.7.5
Sightglass#1408


Innoxia:
Thank you for this! 🆔

PR #1771, opened 2023-05-10
Check feral attributes before setting body as feral

SightglassCat:
Changes the body setFeral method to not set the body as feral if the subspecies does not have feral attributes or is null. Reduces need to check for feral attributes before calling the setFeral method. Still outputs an error message to the log.
Moves setting of the feral boolean after the check for null subspecies or null feral attributes, in the Body.setFeral method.
No new graphical assets.
Tested on versions 0.4.7 to 0.4.8.3.
Sightglass#1408


Innoxia:
Thank you for this! 🎈

PR #1772, opened 2023-05-12
Fix to slime, half-demon backgrounds

SightglassCat:
Fixes missing background textures in slime and half-demon race icons.
Add a new uncolored, icon-only SVG string, which is layered over the slime or half-demon background. Previously, the icon and a background were put over, which covered up the slime splat or demon pentagram background underneath.
No new assets required
Tested on v0.4.8.3
Sightglass#1408


Innoxia:
Thank you for this! 💦

PR #1773, opened 2023-05-13
Added support for Inkscape's Mesh Gradients As well as Reworked Option Buttons and Added Slider Support

DrZed:

 Add support for Inkscape's Mesh Gradients, I have an artist friend who wanted to make clothes for LT using this that later found out it didn't have support for Mesh Gradients and would have to remake everything. So instead I fixed the issue directly by adding it in my personal branch, tested throughly and now pushing to main branch.
 
 EDIT - I also reworked Toggle Buttons, and added Sliders, only converted 2 things, for now, will convert the rest if I get the go ahead from Innoxia
 
 - I changed buttons to now modify their display property instead of redraw the window, only changed them to be consistent with how Sliders are managed. Speaking of sliders, I added sliders support and converted only 2 values to using them for a reference, Will convert the other applicable things to sliders if I get approval. 

 No.

- Has this change been tested? Yes, been running this on my branch of LT for quite a while.  Been in testing for ~4 weeks by ~200 people.

- So we have a better idea of who you are, what is your Discord Handle? KeldonSlayer#6696

### DIFF
--- a/src/com/lilithsthrone/controller/MainController.java
+++ b/src/com/lilithsthrone/controller/MainController.java
@@ -1,6 +1,8 @@
 package com.lilithsthrone.controller;
 
+import java.io.IOException;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.time.LocalDateTime;
 import java.time.format.TextStyle;
 import java.util.ArrayList;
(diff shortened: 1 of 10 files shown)


DrZed:
I see I forgot the negative value fix for sliders that lets it properly display and work with negatives. I'll push that in a moment, after I validate the other 3 spots that use getContentPreferenceVariableDiv()

CognitiveMist:
This sounds like maybe it should be 2 different PRs.

DrZed:
I really should split them, this is quite a late reply

DrZed:
Going to redo this PR as 2 separate ones the day of Innoxia's next push, so I can get it in clean.

PR #1780, opened 2023-06-30
parsing fix in bandit.xml

MintyChip:
Line 206 was missing parsing so I added it.

What about the other 54 additions and 26 deletions? I don't know, github decided to do that itself. And when I tried to push from IntelliJ, it decided the whole document was replaced so I guess this is better??


Innoxia:
I've fixed this manually in my branch as I was hesitant to merge this due to all the formatting changes. I've credited you with the fix in the patch notes. Thank you!

PR #1781, opened 2023-07-06
Sanitize quicksave file name

SightglassCat:
When playing as a character whose name contains a space or period, the quick load button looks for a different filename than what is actually saved.
Changes filename produced by getQuickSaveName to follow the same file name check as when the file is saved.
No
Tested on version v0.4.8.5
sightglass


Innoxia:
Thank you for this!

PR #1788, opened 2023-07-30
Negative RESTING_LUST attribute

SightglassCat:
Negative resting lust bonuses from subspecies or items often have no effect since RESTING_LUST cannot be negative, and they can only affect the resting lust if another source has a positive resting lust modifier.
Changes the RESTING_LUST attribute min/max value to -15/80 from 0/80, and changes the GameCharacter.getRestingLust calculation to ensure that resting lust is always at least 0 (and positive if vibrating equipment is worn).
No
v0.4.8.9
sightglass


Innoxia:
Thank you for this!

PR #1802, opened 2023-10-04
Cache compiled Javascript objects.

tarbh-uisge:
This makes significant improvements to the speed of UtilsText.evaluate - in my measurement, with a save with around a hundred children and a bunch of script-heavy mods installed, it cut about 60% of the _script evaluation_ time out of resting until the following morning. Sadly, this only amounts to a few percent of the overall time spent working out resting, but every bit helps.

This improvement also applies to the multitude of other situations where scripts are run, incrementally improving performance of walking around, opening the inventory, enchanting and dying objects, sex and combat, and numerous other areas of the game. It's not a _huge_ improvement, but it's noticeable.

This implementation is exactly the implementation that was removed in 81a1fc0b14 (version 4.1.0), with one key difference: the cache of compiled scripts is cleared whenever the script engine is changed or cleared. The issues you were having with it sometimes failing to run scripts were _likely_ due to a compiled script object carrying over a save reload or other situation where the script engine gets discarded, leading to scripts being run in engine A while variables were being set in engine B.

I've been running with this patch locally for a while and it seems stable.

You can reach me on Discord as `tarbh_uisge`.

### DIFF
--- a/src/com/lilithsthrone/game/dialogue/utils/UtilText.java
+++ b/src/com/lilithsthrone/game/dialogue/utils/UtilText.java
@@ -9777,6 +9777,7 @@ public class UtilText {
 	
 	public static void resetParsingEngine() {
 		engine = null;
+		memo.clear();
 		specialParsingStrings = new ArrayList<>();
 	}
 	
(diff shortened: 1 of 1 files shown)


Innoxia:
Thank you for this! I've added you to the developers channel on the LT discord.

PR #1812, opened 2024-02-07
removed most slurs

Fghwe44:

remove most of the slurs, this is not a perfect solution nor am i the arbiter of all slurs, these are just those that i recognize as the most in need of change.

i would really like to see this changed as this game has helped me on my gender discovery journey and id love to recommend this game to more people but nearly all of them are put off by the usage of slurs in the game

replaced slurs with less offensive words

no

no, as i do not have sufficient experience with the IDE's that build guides exist for nor do i have enough experience with java to try and figure it out, since this is a simple renaming i hope that this is acceptable in this case 

"thestupidsage"

### DIFF
--- a/src/com/lilithsthrone/game/character/CharacterUtils.java
+++ b/src/com/lilithsthrone/game/character/CharacterUtils.java
@@ -2125,7 +2125,7 @@ public class CharacterUtils {
 			} else {
 				character.setPenisVirgin(true);
 			}
-			if((character.getGender()==Gender.F_P_TRAP || character.getGender()==Gender.N_P_TRAP)) {
+			if((character.getGender()==Gender.F_P_FEMBOY || character.getGender()==Gender.N_P_NONBINARY_FEMBOY)) {
 				float sizeAlteration = (100+Main.getProperties().trapPenisSizePreference)/100f;
 
 				character.setPenisSize(character.getPenisSize().getMinimumValue() + Util.random.nextInt(character.getPenisSize().getMaximumValue() - character.getPenisSize().getMinimumValue()) +1);
--- a/src/com/lilithsthrone/game/character/GameCharacter.java
+++ b/src/com/lilithsthrone/game/character/GameCharacter.java
@@ -24796,7 +24796,7 @@ public abstract class GameCharacter implements XMLSaving {
 							isPlayer()
 							?"Due to the fact that both your [pc.vagina] and [pc.penis] are exposed, as well as the fact that you have [pc.breastSize] breasts, everyone can tell that you're [pc.a_gender("+colouredGender+")] on first glance."
 							:"Due to the fact that both [npc.her] [npc.vagina] and [npc.penis] are exposed, as well as the fact that [npc.sheHasFull] [npc.breastSize] breasts, everyone can tell that [npc.sheIs] [npc.a_gender("+colouredGender+")] on first glance.",
-							Gender.F_P_V_B_FUTANARI);
+							Gender.F_P_V_B_FEMININE_INTERSEX_BOOBS);
 						
 				} else if(visibleVagina) {
 					// Exposed vagina and obvious penis bulge:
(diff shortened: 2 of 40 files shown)


Moth-GITHUB:
Heyo
Not to be rude or anything but I highly doubt this PR will be merged based on 2 reasons.
1: all these "slurs" can be changed to "non-slurs" in the options menu of the game in the "gender pronouns" options
2: for simplicity's sake the variables are such because I think the dev finds it easier and quicker to write "F_P_V_B_FUTANARI" instead of "F_P_V_B_FEMININE_INTERSEX_BOOBS"

and also just an honest question because I'm curious, I know this probably isn't the right place to ask but why have you changed "Trap" to "femboy" considering they're not even related. based on my (how I personally see it) limited conservative leaning knowledge.

No harm or offence meant, I'm glad you're concerned and want to change the "slurs" to something else
Just making a friendly comment/question

Fghwe44:
TLDR as for why trap is a slur:
It implies that people that have non matching genitals to their gender representation want to trap people to deceive them into having sex with them and this is problematic in general (google “trans panic defence” for why)

The longer version:
https://intoomanywords.com/2017/12/05/why-trap-is-a-bad-term/

To 1. Yes, it is good to have that menu but it’s also majorly offputting for people in the minorities those slurs target and they shouldn’t be there in the first place as this continues their usage


To 2. Must modern IDE’s have auto complete and since the starting letters basically already tell you what the term after it should be this shouldn’t present an issue

Moth-GITHUB:
Fair enough. I understand your points.
I will not interact further as there is no need to waste both our times in arguing with counter arguments.
I will just say that it will be kind of hard to replace the word trap due to the context the word is majorly used in.
and also I feel like "femboy" isn't the right replacement for the context that word is used in the game.
I hope you have a good-day/night and I wish you good fortune in trying to make the world a better place.

CognitiveMist:
I have three mostly-independent comments to make:

---

If I may nitpick the methodology here, using one example:

The structure of an identifier such as `F_P_V_B_FUTANARI` already includes the information that it's a feminine body type, with a penis, vagina, and breasts. The name after (futanari) is merely quick easily-readable check to make sure that you've got the right thing (as opposed to some piece of clothing, or an enchantment effect).

So, to avoid the word 'futanari', you don't really need `F_P_V_B_FEMININE_INTERSEX_BOOBS`, since that's mostly redundant. `F_P_V_B_INTERSEX` would do the job equally well. (Actually, better, imo. Yes, there is tab completion, but excessively long names hurt readability and increase congitive load).

---

This simple rename is unlikely to be accepted as-is because it may break existing mods and save files. These are internal identifiers, and as such, they are used within mods and save files as a more human-readable way than just some numbers. Useful for debugging.

This is not unresolvable, but it does mean that some extra logic would have to be added to watch for the old IDs and map them over to the new IDs where necessary. For save files, that's pretty straightforward, but mods might be considerably less so.

---

I find it a bit odd that you'd change these identifiers which will never be seen when playing the game, but not change the default values for the gender names options menu in `GenderNames.java`. From your initial PR description, it sounds like you've never actually compiled and run this modified version, so you may not realize that these changed identifiers are not what designate those default values.

Hulkninja:
Words change meaning.

I've never in my entire life heard of someone saying "you f*cking trap" to someone in order to insult or hurt them.

Fact is, where I'm from, "Polack" is the standard, non-derogatory word for Polish people. 

That's because "Polack" became a slur, but only to some people, based on language, geography, etc. 

Do you see where I'm going with this?

Words are inherently subjective.

Words are also constantly evolving organically, and trying to manually control language is an authoritarian thing to do by it's very nature.

I'll keep saying trap, because it means "person with a penis, testicles, and an androgynous appearance".

If you want that word to mean something else to you, that's your decision.

If the game is changed then that's cool by me but this feels less like a game dev thing and more like an ideological thing.

Moth-GITHUB:
> Words change meaning.
> 
> I've never in my entire life heard of someone saying "you f*cking trap" to someone in order to insult or hurt them.
> 
> Fact is, where I'm from, "Polack" is the standard, non-derogatory word for Polish people.
> 
> That's because "Polack" became a slur, but only to some people, based on language, geography, etc.
> 
> Do you see where I'm going with this?
> 
> Words are inherently subjective.
> 
> Words are also constantly evolving organically, and trying to manually control language is an authoritarian thing to do by it's very nature.
> 
> I'll keep saying trap, because it means "person with a penis, testicles, and an androgynous appearance".
> 
> If you want that word to mean something else to you, that's your decision.
> 
> If the game is changed then that's cool by me but this feels less like a game dev thing and more like an ideological thing.

Man I was surprised to see that word in my notifications
(refering to "Polack") I know I said I won't get involved in this discussion anymore but I have nothing better to do.
all I just wanted to say is I don't get that word. and I'm polish myself, I head it was like a "slur" to polish people but I mean like to be frank I feel like it's just non-polish speakers taking words and giving them a bad meaning to be cruel to others because in polish you say "Polak" which literarlly just means someone who's polish. 

Also on your definition of a trap it makes sense but traps can also be female. from the way I learned that term it was mostly refering to an anime character who looks like a female but is male and vice-versa (however some people add "reverse" to trap if it's female that looks like male)

anyway have a goodday/relevant timezone equivlant

PR #1813, opened 2024-02-20
Documentation: Specify svgo version compatible with svgoConfig.yml

CognitiveMist:
Newer versions of svgo don't support yaml configuration, and also no longer support some features enabled within this config (at least, that was true when I last checked like a year ago).

The pinned version (1.3.2) is the version I have personally used, and also is the most recent version that still supports yaml.

Discord: `@phlarx`


Innoxia:
Thank you for this!

Issue #1819, opened 2024-03-27
Orgasm without fluid still stains clothing under certain conditions


alfadorfox:
**Describe the bug**
When a character has no penis or vagina, the orgasm produces no fluid. However, if the orgasm is clothed (via self-stimulation of other body parts such as breasts) then the clothing covering the groin may still be flagged as dirty. This is because the check for penis orgasm cum quantity in the generic orgasm effect does not check for the presence of a non-NONE penis, only that the testicles of the character's body produce cum, and a newly generated character will always have nonzero testicle production numbers, even if the character generates as female and thus has no penis.

https://github.com/Innoxia/liliths-throne-public/blob/dev/src/com/lilithsthrone/game/sex/sexActions/baseActionsMisc/GenericOrgasms.java#L2633
https://github.com/Innoxia/liliths-throne-public/blob/dev/src/com/lilithsthrone/game/sex/sexActions/baseActionsMisc/GenericOrgasms.java#L8308

**Expected behavior**
Expected behavior is that a character with no penis should not be staining clothing with semen. This may have gone unnoticed if the only testing done of clothed masturbation was done with characters who have either a penis or vagina.

**To Reproduce**
1. Start a new game and create a Female character.
2. Skip tutorial
3. Rest 12 hours so shops will be open
4. Purchase 14 arcane essences and absorb them
5. Unlock enchanting and enchant any clothing item to remove vagina in one hour
6. Wait one hour until vagina disappears
7. Masturbate via cell phone
8. Use grope breasts (self) action repeatedly until orgasm
9. Observe that the text describes a fluid-free orgasm, yet the groin slot item equipped (or legs if no groin item is equipped) will become dirty.

The same occurs if you begin as a Male character and use an enchantment to remove your penis, since removing the penis does not alter the testicle stats.

**Attach files**
not necessary as this is not a bug that causes an error to be logged

**Additional information:**
 - Game version [e.g. 0.3.2.5] v0.4.9
 - OS: [e.g. Windows 10] Windows 10
 - Using .jar or .exe: [e.g. .jar] .exe
- If using .jar, and you are not including an error.log file/text, please include your java version: [e.g. 1.8.0_172] n/a

PR #1820, opened 2024-03-27
Fix Issue #1819 - check for a non-dildo penis when staining due to clothed orgasm

alfadorfox:
Quick fix for issue #1819 

Does not impact clothing staining for characters with vaginas, since that is handled by separate squirting logic in the Sex class.


Innoxia:
Thank you for this! If you use discord and are on the LT server, then please let me know what your discord username is and I'll add you to the developers channel there.

PR #1823, opened 2024-04-03
Bugfix: Prevent infinite loop when computing the visible clothing for a slot

CognitiveMist:

Determining the visible clothing for a slot has a possibility of getting stuck in an infinite loop.

In this change, the loop is prevented by:
1. using a set to grab the slots to scan
2. using a second set to keep track of already-visited slots

Also, stored `getInventorySlotsConcealed` to a var, so that it's only called once
I'm fairly sure there won't be any differences to the original implementation, in the results that are generated, but it's not impossible.

However, I also attempted to rewrite the entire function from scratch, though my implementation will produce differing results. Since the differences tend to imply that there's code in other functions that might need to change, too, I have not included that implementation here. It is visible here: https://github.com/innoxia/liliths-throne-public/compare/dev...CognitiveMist:visible-clothing-freeze-rewrite .

Nope

Yep, 0.4.9.5

`phlarx` (Cognitive Mist)

### DIFF
--- a/src/com/lilithsthrone/game/inventory/CharacterInventory.java
+++ b/src/com/lilithsthrone/game/inventory/CharacterInventory.java
@@ -1112,24 +1112,29 @@ public class CharacterInventory implements XMLSaving {
 			visibleClothing.add(getClothingInSlot(slot));
 		}
 		
-		if(getInventorySlotsConcealed(character).get(slot)!=null) {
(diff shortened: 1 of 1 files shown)


CognitiveMist:
Self-closing because this was fixed in almost the same way on `dev`.

PR #1829, opened 2024-05-02
Enhancement: Add sybaritic potion (refractory potion)

CognitiveMist:
- Consumable potion that increases an npc's desired orgasms by 3 - 5 during sex.
- Added map in Sex to keep track of the bonus.
- Map can also handle decreases, limiting to a minimum of 1 orgasm, but this is currently unused. There are currently no limits for the value within the map.
- Out of sex, consuming the potion will give a few essences.
- In sex, it will have no tangible affect on the player, since the player doesn't use the minimum-orgasms mechanic.
- ID is distance 6 from the mod preview: `cog_refractory_refractory` -> `phlarx_refractory_refractory`, so it should always convert items correctly if they uninstall the mod. (I will remove the mod post when this is merged)

image
image

Of course, feel free to tweak / tune as you see fit.

Developed in version 0.4.9.9

Discord: `phlarx`


Innoxia:
Thank you for this!

PR #1831, opened 2024-05-30
Bugfix: Minor fixes from recent discord bug report

CognitiveMist:
- Convert `pregnancyReactions` to a Set
- Change `continue` to `break` in one place
- Prevent generating leg configuration TF text for a character that is null (via `Body.applyLegConfigurationTransformation`)
- Fix one case of incorrect SPECIAL_PARSE_0 usage (this is the only example I found within the code base)

Original bug report info (not actually a bug): https://discord.com/channels/302617912949211136/789126893658963988/1245837478485823612
All items discovered by me while investigating that report.

Tested 0.4.9.9, `@phlarx`


Innoxia:
Thank you for this!

PR #1839, opened 2024-12-10
Fix Eisek (and probably other NPCs) showing up as winged dragon-boy

Stadler76:
  Fixed Eisek (and probably other named NPCs) showing up as 'The **winged** dragon-boy'.

  Eisek (and probably other NPCs) showed up as 'The winged dragon-boy' because getFleshSubspecies() always returned 'HUMAN' for him for whatever reason. I've adjusted the code to only check for getFleshsubspecies() if the BodyMaterial is not flesh and for getSubspecies() otherwise.

  Did a few tests with Eisek, my occupants and some slimes including the ones in the Slime Queen's Tower and everything works as intended so far.

  No

  Yes, 0.4.10.7 Alpha

  `Stadler#3007`


Innoxia:
Thank you for this!

PR #1840, opened 2024-12-14
Fix logic error in SlaveJob

SquishyPandaDev:
Fix the player being unable to assign the maximum number of slaves to certain jobs.

Discord: mightbefluffy

### DIFF
--- a/src/com/lilithsthrone/game/occupantManagement/slave/SlaveJob.java
+++ b/src/com/lilithsthrone/game/occupantManagement/slave/SlaveJob.java
@@ -595,7 +595,7 @@ public enum SlaveJob {
 		public boolean isAvailable(int hour, GameCharacter character) {
 			return !character.getHomeLocationPlace().getPlaceType().equals(PlaceType.SLAVER_ALLEY_SLAVERY_ADMINISTRATION)
 					&& !character.getHomeWorldLocation().equals(WorldType.getWorldTypeFromId("innoxia_dominion_sex_shop"))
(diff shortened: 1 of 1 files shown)


CognitiveMist:
I'm not sure that this changes the result in a meaningful way, since `a < b` and `!(a >= b)` are completely equivalent for integers.

[Here's some example code to demonstrate.](https://tio.run/##PY5NDoIwEIXX5RQDqzZENsa44OcGrlgaF1NQUuUvtJAYw9YDeEQvUqdIaJq@zjfvTeaOE@7u5cPafpS1KqCoUWs4oWrh5bEVaoOGZOpUCQ21eG4G1VbnC@BQaeGcTLUGEFI4xFTcuoE7IAnsY5IkhSNpGP7NLH9qc22ibjRRT6NM3XKHGUIIQRLQK90Pvu8PuIpjIsVC1kNwCQQ@d/0ll6VbUGxJn2OWSuHMwq02e3Rna38)

SquishyPandaDev:
Oh dear I made a logic mistake too. And I was over thinking it too _face palm_. Sorry about that

The problem is this. Lets say you have 7 slaves assigned to milking and you go to add a 8th. That 8th slave will show as having an unavailable time slot. The function as is will get the current number of slaves in that job slot (8) and compare it to the max of 8. 8 < 8 returns false. No free slots.

It should be a <= b. I will change it and resubmit

Innoxia:
Thank you for your PR, but I've fixed this bug by adding a check to see if the character is already working the job at the specified hour, which should sort it out.

PR #1842, opened 2024-12-18
Add support for chastity items with an attached dildo

SquishyPandaDev:
Tried to mod in a chastity cage with a dildo and found that the game logic will pick the characters penis. Changed __getCurrentPenis__ to always check clothing for a dildo. And now the dildo is selected.

I don't know what else needs to be changed to further support this. If this is something that sounds interesting to put in the game, I'm willing to make the needed code changes.

Discord: mightbefluffy

### DIFF
--- a/src/com/lilithsthrone/game/character/GameCharacter.java
+++ b/src/com/lilithsthrone/game/character/GameCharacter.java
@@ -30248,20 +30248,41 @@ public abstract class GameCharacter implements XMLSaving {
 			return null;
 		}
 		
-		if(body.getPenis().getType()==PenisType.NONE) {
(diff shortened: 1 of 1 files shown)


SquishyPandaDev:
I checked the item tag PREVENTS_ERECTION_PHYSICAL to avoid having to iterate over another loop. Will that do

SquishyPandaDev:
I used the item tags PREVENTS_ERECTION_PHYSICAL and CHASTITY to check if penis is blocked by a cage. Did this to avoid having to iterate over another list.

SquishyPandaDev:
There is a bug when a character orgasm. Will fix

PR #1844, opened 2025-01-13
Multithreaded Preloading of Data

DrZed:
    - To improve Save Loading speed (up to 440% faster)
    - To improve New Game speed (up to 104% faster)

    - Added DocBuilders.java
         - Replaces the Document Builder in Main with a dedicated cluster
         - This was required to make threading work, as JAXP is not thread-safe
     - Added PreInitializationThread.java
         - This is called fairly early in the stack, before the window is requested to even draw
         - Doing this allows things such as BodyCoverings, Race, ETC to be already prepared when needed
     - Added NPCRegistry.java
         - This moves NPC's from Game.java and makes it easier to see/sort/modify them
         - Split Relationship Registration to it's own function below the NPC Initialization
     - Added NPCThread.java
         - This allows NPC's to be initialized in clusters on separate threads

    - Not at all

   - This change has been tested and proven for 4 years in my own build mod.
   - It has finally been fully vetted as stable, by roughly 5-11k users varies by version number.

    - @keldonslayer

    Done did.

### DIFF
--- a/license.md
+++ b/license.md
@@ -14,6 +14,7 @@ Although this Project is publicly available, it is not open source software. All
 
 "Contribution" shall be defined as submitting any Content, idea, or suggestion available to the Project creator(s) and/or maintainers by any means, including e-mail, file transfer protocols of any kind, git pull requests, git pushes, or any other electronic or physical transfer of any kind which makes available to the project creator(s) and/or maintainers for inclusion in the Project.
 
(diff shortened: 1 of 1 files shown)


DrZed:
I can split these 2 commits into separate PR's if wanted, however both depend on DocBuilders due to the nature of JAXP.

PR #1850, opened 2025-03-16
Mod support: combatAction blockAmount

miome:
Currently, if a mod implements a CombatAction, and uses the blockAmount tag, the blockAmount tag is ignored.

This request has AbstractCombatMove check if the combat move running is a mod, and if blockAmount > 0, and if so applies blockAmount to shielding.  

The shielding is always applied to the source character, even if there is a target.  I'm not sure if blockAmount is meant to support applying shielding to an ally, but I didn't see any indication it could.  If the mod should be able to target allies. I have some ideas on how to check for that, though, if blocking for allies is needed.


miome:
have a stray print statement in here, will re-submit without it

PR #1862, opened 2026-02-19
Development made easier

lure0xaos:
@Innoxia 
**You could get some rework of LT**

Detailed description is available in [release](https://github.com/lure0xaos/liliths-legacy/releases/tag/Demo)


N3X15:
TBH, each atomic change (gradle, webfiew, tooltips, close confirm) should probably be moved to their own PRs in case Inno says no to any of them.

OTOH, I don't really see any reason any of these would get turned down.

lure0xaos:
> TBH, each atomic change (gradle, webfiew, tooltips, close confirm) should probably be moved to their own PRs in case Inno says no to any of them.
> 
> OTOH, I don't really see any reason any of these would get turned down.

That's true for closing window feature but other should be one because unclickable buttons on the left and right mean the game is mostly unplayable (though hotkeys are still usable). 

Oh and I should mention that all import java.io.* and java.nio.* as well as javax.xml.* should be replaced by import com.lilithsthrone.utils.* respectively (see diffs) to work with VFS, in any further development. So many changes are new files and many import changes. Also, com.lilithsthrone.res.* resources are moved from src ro srcres to separated source code and any binary files (xml, images etc).
Also, I've should separate Gradle, VFS and logging in case Inno wouldn't like any of this, but it seems to me it it logical to have all of these as one.
Any other improvements will go to separate commits as well separate PR, and some of these will go to separate branches that have main branch as a base.

CognitiveMist:
All of the files that are just changing the order of imports produce unnecessary noise.

Unless there is a specific benefit to such changes, they should just adhere to the existing project style.

PR #1869, opened 2026-09-10
Birthday to 1-31 days

DrZed:
Changed Birthday's for any NPC with random birthday to do 1-31 days, then clamping it to the birth month's day limit in NPC

Change birthday to be able to be any day of a month, I found it weird the days were never 29th/30th

Changed all 1+Util.random.nextInt(25) / 1+Util.random.nextInt(26) / 1+Util.random.nextInt(27) in birthday declarations for NPCs to 1+Util.random.nextInt(30), 
and in NPC.java changed (birthMonth == Month.FEBRUARY && birthDay==29?28:birthDay) to Math.min(birthMonth.maxLength(), birthDay)

nada

Yes, 0.4.11

@keldonslayer

### DIFF
--- a/src/com/lilithsthrone/game/character/npc/NPC.java
+++ b/src/com/lilithsthrone/game/character/npc/NPC.java
@@ -166,8 +166,8 @@ public abstract class NPC extends GameCharacter implements XMLSaving {
 			NPCGenerationFlag... generationFlags) {
 		super(nameTriplet, surname, description, level,
 				age<MINIMUM_AGE
-					?LocalDateTime.of(Main.game.getStartingDate().getYear()-age, birthMonth, (birthMonth==Month.FEBRUARY&&birthDay==29?28:birthDay), 12, 0)
-					:LocalDateTime.of(Main.game.getStartingDate().getYear()-(age-MINIMUM_AGE), birthMonth, (birthMonth==Month.FEBRUARY&&birthDay==29?28:birthDay), 12, 0),
+					?LocalDateTime.of(Main.game.getStartingDate().getYear()-age, birthMonth, Math.min(birthMonth.maxLength(), birthDay), 12, 0)// Why not random hour/minutes?
+					:LocalDateTime.of(Main.game.getStartingDate().getYear()-(age-MINIMUM_AGE), birthMonth, Math.min(birthMonth.maxLength(), birthDay), 12, 0),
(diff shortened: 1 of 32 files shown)


src/com/lilithsthrone/game/character/npc/NPC.java
```diff
@@ -165,11 +164,10 @@ protected NPC(boolean isImported,
...
 			boolean addedToContacts,
 			NPCGenerationFlag... generationFlags) {
-		super(isImported,
-				nameTriplet, surname, description, level,
+		super(nameTriplet, surname, description, level,
 				age<MINIMUM_AGE
-					?LocalDateTime.of(Main.game.getStartingDate().getYear()-age, birthMonth, (birthMonth==Month.FEBRUARY&&birthDay==29?28:birthDay), 12, 0)
-					:LocalDateTime.of(Main.game.getStartingDate().getYear()-(age-MINIMUM_AGE), birthMonth, (birthMonth==Month.FEBRUARY&&birthDay==29?28:birthDay), 12, 0),
+					?LocalDateTime.of(Main.game.getStartingDate().getYear()-age, birthMonth, Math.min(birthMonth.maxLength(), birthDay), 12, 0)// Why not random hour/minutes?
+					:LocalDateTime.of(Main.game.getStartingDate().getYear()-(age-MINIMUM_AGE), birthMonth, Math.min(birthMonth.maxLength(), birthDay), 12, 0),
```
CognitiveMist, line 170: Since the month and day are just discarded, I feel like it would be a simpler system to take the age and a number of days between 0 and 364, then do the following instead of the whole `LocalDateTime.of( ... )`
```java
Main.game.getStartingDate().minusYears(age).minusDays(days)
```

Thats a uniform distribution across all days except Dec 31 of leap years.

You could do `minusMonths` as well, but it changes the meaning of the month value. Instead of saying which month of the year was their birthday, it's saying how many months ago was their birthday. And it would give an inferior distribution anyhow.

---

To answer the question in the comment: the hour and minute are never used by the game, and they're not stored in the save file, so doing so would have no benefit without also addressing those two points.

DrZed: Good call about the minus days days, then I guess I'll need to add an additional constructor to NPC.java to support the Uniques having specific birthdays.

Also jesus christ github, only notifying me now that there was a comment?

